Issue #4, opened 2022-02-09
Improve cross-compilation example


lovesegfault:
Currently, the cross-compilation example instantiates `nixpkgs` like so:

https://github.com/ipetkov/crane/blob/5afc3e94fdf7731ccea2395947f1cd37aec5a2d5/examples/custom-toolchain/flake.nix#L23-L26

And proceeds as usual, only referencing the custom toolchain when instantiating `craneLib`:

https://github.com/ipetkov/crane/blob/5afc3e94fdf7731ccea2395947f1cd37aec5a2d5/examples/custom-toolchain/flake.nix#L28-L40

This is AOK if you don't have any system dependencies, but it would be great if an example were added showing how to use `crane` together with `crossSystem`, akin to the example in rust-overlay: https://github.com/oxalica/rust-overlay/blob/master/examples/cross-aarch64/shell.nix.

I've attempted to get this to work with the following (truncated) snippet:
```Nix
{
        localSystem = "x86_64-linux";
        crossSystem = "aarch64-linux";
        pkgs = import nixpkgs {
          inherit localSystem crossSystem;
          overlays = [ rust-overlay.overlay ];
        };

        rustToolchain = pkgs.pkgsBuildHost.rust-bin.stable.latest.default;

        craneLib = (crane.mkLib pkgs).overrideScope' (final: prev: {
          cargo = rustToolchain;
          rustc = rustToolchain;
          clippy = rustToolchain;
          rustfmt = rustToolchain;
        });

        src =  ./.;

        cargoArtifacts = craneLib.buildDepsOnly {
          inherit src;
        };

        crateFmt = craneLib.cargoFmt {
          inherit cargoArtifacts src;
        };

        crateClippy = craneLib.cargoClippy {
          inherit src;
          cargoArtifacts = crateFmt;
          cargoClippyExtraArgs = "-- --deny warnings";
        };

        crate = craneLib.buildPackage {
          inherit src;
          cargoArtifacts = crateClippy;
        };
}
```

But it fails with the following:
```
hyperpixel_init-clippy-aarch64-unknown-linux-gnu> unable to copy cargo artifacts, "/nix/store/ppzyixshipwzr7zqvq190mdfbcap6njg-hyperpixel_init-fmt-aarch64-unknown-linux-gnu-0.1.0/target.tar.zst" looks invalid
```

You should be able to reproduce this with `nix build -L` on https://github.com/lovesegfault/hyperpixel-init on the `crane` branch.

[...]

PR #5, opened 2022-02-09
fix(pkgs): use zstd from pkgsBuildBuild

lovesegfault:
This allows for crane to work when `buildPlatform != hostPlatform`, such
as well `crossSystem` is set.

This partially addresses #4

### DIFF
--- a/pkgs/default.nix
+++ b/pkgs/default.nix
@@ -2,7 +2,7 @@
 , makeSetupHook
 , jq
 , rsync
-, zstd
+, pkgsBuildBuild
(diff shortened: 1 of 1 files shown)


ipetkov:
Ah, thanks for addressing this! I'm not very well versed with cross compiling under nix, but this change makes sense.

Wonder if it would be useful to scope other tools in this manner (e.g. rustfmt, cargo-tarpaulin, etc.) since they only need to run for the `buildPlatform`... Will leave it for a future PR

PR #18, opened 2022-03-14
fix(lib/buildDepsOnly): don't check for Cargo.lock if user configures lock path

lovesegfault:
This should allow using buildDepsOnly with a custom lockfile path.

### DIFF
--- a/lib/buildDepsOnly.nix
+++ b/lib/buildDepsOnly.nix
@@ -28,7 +28,7 @@ let
 
   path = args.src or throwMsg;
   cargoToml = path + "/Cargo.toml";
-  cargoLock = path + "/Cargo.lock";
+  cargoLock = args.cargoLock or (path + "/Cargo.lock");
   dummySrc =
     if builtins.pathExists cargoToml && builtins.pathExists cargoLock
     then mkDummySrc args


ipetkov:
Thanks for the PR @lovesegfault!

I had originally considered making `buildDepsOnly` able to take a custom Cargo.toml or Cargo.lock but decided against it to avoid an explosion of various options that would be accepted (at least until I could better understand the specific use-cases that need it). `buildDepsOnly` will pretty much delegate to `mkDummySrc` which will also crawl the `src` input for any `.cargo/config.toml` files along with consuming Cargo.toml and Cargo.lock.

In my mind, I imagined that if a custom Cargo.lock is necessary, then the caller should copy it (along with Cargo.toml) into its own directory and then invoke `buildDepsOnly` with that result. This gives the caller the maximum flexibility to configure things as they wish without introducing too many knobs to dial on the function itself (which I would be happy to improve the docs on for future readers).

I am curious to learn what your use-case is, and if the suggestion above is a valid solution for it?

yusdacra:
I have also come to need this (over on dream2nix). I used a separate derivation to copy the Cargo.lock, but that seems to trigger IFD (`nix flake check` would fail etc.). So I would prefer if this PR was chosen as the solution. It maybe that I was doing something wrong (was just using a simple `runCommandLocal` call...), but that's what I experienced.

And FYI the use case here is that some sources (be it fetched remotely or local) may not have `Cargo.lock` files. `dream2nix` generates a `Cargo.lock` file from it's `dream-lock.json` and puts it in the dummy source using the `cargoLock` argument. This does not trigger IFD because the Cargo.lock can be generated fine on the fly and be made into a path (`builtins.toFile` doesn't trigger IFD anymore, not on stable yet but it is coming :tm:). For the actual `buildPackage` derivation, it just uses a `cat ${cargoLock} > Cargo.lock`. That's how we use it, if it helps.

ipetkov:
> I have also come to need this (over on dream2nix). I used a separate derivation to copy the Cargo.lock, but that seems to trigger IFD (`nix flake check` would fail etc.)

@yusdacra What is the exact IFD error you are hitting? IIUC IFD is always enabled when running as a flake (since all inputs are pure by default), though I have seen some weirdness occur when the flake defines multiple systems.

---

> And FYI the use case here is that some sources (be it fetched remotely or local) may not have `Cargo.lock` files.

`vendorCargoDeps` is the only place where the contents of `Cargo.lock` are consumed; by default crane will inspect the contents of the file (effectively IFD if the path isn't local) as a convenient default for common projects using flakes. It's entirely possible to invoke it directly and side step all the headache:

```nix
let
  cargoLock = ./path/to/locally/written/Cargo.lock;
  src = fetchUrl whatever; # Path to remote source
in
craneLib.buildPackage {
  inherit src;

  # Explicitly vendor dependencies from our local Cargo.lock
  cargoVendorDir = craneLib.vendorCargoDeps {
    inherit cargoLock;
  };

  patchPhase = ''
    # Make our locak Cargo.lock available to the build
    cp ${cargoLock} Cargo.lock
  '';
}
```

ipetkov:
Taking a closer look at this, #32 basically got rid of the need for this (by loosening how we copy `Cargo.lock` files, i.e. do nothing if the Cargo.lock file is missing and leave it up to the caller to worry about providing one). I just removed the check altogether in #39 so I believe any issues this was causing should be gone!

I'm going to close this PR out as it's no longer needed, but if I'm missing something please feel free to chime in!

yusdacra:
Yes, the issue is indeed fixed, thanks!

About the IFD issue, it was just that the evaluation of `nix flake check` was failing since it tries to eval all systems, but this of course fails since there are no builders for those systems.

PR #27, opened 2022-05-11
flake: fix cross-musl path

TristanCacqueray:


ipetkov (approved):
Whoops, great catch @TristanCacqueray !

PR #32, opened 2022-05-17
mkDummySrc: fix cache invalidation if src is flake root

### LINKED ISSUE #28: Dummy source derivation relying on source when using flake's self argument as source
When using `self` over `./.` for the source to `buildPackage`, the dummy source derivation relies on the source of the flake.

Good:
```json
    "inputSrcs": [
      "/nix/store/xxx-default-builder.sh",
      "/nix/store/xxx-Cargo.lock"
    ],
    "inputDrvs": {
      "/nix/store/xxx-dummy.rs.drv": [
        "out"
      ],
      "/nix/store/xxx-Cargo.toml.drv": [
        "out"
      ],
      "/nix/store/xxx-stdenv-darwin.drv": [
        "out"
      ],
      "/nix/store/xxx-bash-5.1-p16.drv": [
        "out"
      ]
    },
```

Bad:
```json
    "inputSrcs": [
      "/nix/store/xxx-default-builder.sh",
      "/nix/store/xxx-source"
    ],
    "inputDrvs": {
      "/nix/store/xxx-bash-5.1-p16.drv": [
        "out"
      ],
      "/nix/store/xxx-stdenv-linux.drv": [
        "out"
      ],
      "/nix/store/xxx-Cargo.toml.drv": [
        "out"
      ],
      "/nix/store/xxx-dummy.rs.drv": [
        "out"
      ]
    },
```

A reproduction is available at https://github.com/winterqt/crane-dummy-src-invalidation. This may be outside of Crane's control (as in, this may be an issue with how this attribute of flakes are implemented).

ipetkov:
Fixes #28


iwinux:
Would you consider merging this?

iwinux:
Would you consider merging this?

ipetkov:
@iwinux unfortunately this doesn't end up resolving the originally reported issue and I'm still figuring out why :/

ipetkov:
Alright I've figured out the issue!

Turns out when coercing a derivation to a string Nix will keep track of a hidden "context" of the derivation itself. Any other strings resulting from operations on that path also inherit the same context, so even though we were trying to be clever and copy dummy data around at evaluation time, Nix would still see through it and consider the original source as an input (if it wasn't a relative path, that is).

I did find the `builtins.unsafeDiscardStringContext` which allows us to explicitly _break_ this dependency solving the issue :sunglasses:

PR #40, opened 2022-06-05
fix(mkDummySrc): don't check if cargo lock path exists

yusdacra:
This removes an IFD if the cargo lock path is the output of a derivation. The check for this is useless IMO, as the build will fail anyways if the specified path doesn't exist. This change allows dream2nix to use crane without triggering any IFD (when building crates that are local / flake sources).


ipetkov (approved):
Looks good, thanks!

PR #47, opened 2022-06-30
mkDummySrc: fix handling when src is already filtered

### LINKED ISSUE #46: mkDummySrc doesn't behave well with manually filtered source
Using the following source filter as the argument to `crane.buildPackage`:
```nix
lib.cleanSourceWith {
  inherit src;
  filter = path: type: lib.hasPrefix "Cargo" (builtins.baseNameOf path) || type == "directory";
}
```
results in the generated dummy source only containing `Cargo.toml`.

I don't see anything in `mkDummySrc` that would be causing this behavior, and my filter is sane (I manually tested to make sure it contains both `Cargo.toml` and `Cargo.lock`). My Crane input is at https://github.com/ipetkov/crane/commit/223aaedd36706f0a6fea8080a47f6b5e3e018ede. I've also confirmed that the given source works when not filtering.

ipetkov:
Fixes #46

### DIFF
--- a/lib/mkDummySrc.nix
+++ b/lib/mkDummySrc.nix
@@ -78,17 +78,25 @@ let
   # directory, we check if it happens to be an ancestor for an interesting file (i.e. is a prefix of
   # an interesting file). That way we are left with the smallest possible source needed for our
   # dummy derivation, and we bring any cache invalidation to a minimum. Whew!
-  mkBasePath = p: (toString p) + "/";
-  uncleanSrcBasePath = mkBasePath src;
 
-  uncleanFiles = findCargoFiles src;
+  # NB: if the `src` we were provided was filtered, make sure that we crawl the `origSrc`! Otherwise
+  # when we try to crawl the source Nix will evaluate the filter(s) fully resulting in a store path
+  # whose prefix won't match the paths we observe when we try to clean the source a bit further down
+  # (Nix optimizes multiple filters by running them all once against the original source).
+  # https://github.com/ipetkov/crane/issues/46
+  origSrc =
+    if src ? _isLibCleanSourceWith
+    then src.origSrc
+    else src;
+
+  uncleanSrcBasePath = (toString origSrc) + "/";
+  uncleanFiles = findCargoFiles origSrc;
 
   cargoTomlsBase = uncleanSrcBasePath;
   inherit (uncleanFiles) cargoTomls;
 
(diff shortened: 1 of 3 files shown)


lib/mkDummySrc.nix
```diff
@@ -78,17 +78,25 @@ let
...
+
+  uncleanSrcBasePath = (toString origSrc) + "/";
+  uncleanFiles = findCargoFiles origSrc;
 
   cargoTomlsBase = uncleanSrcBasePath;
   inherit (uncleanFiles) cargoTomls;
 
   cleanSrc =
     let
-      adjustPaths = builtins.map (p: removePrefix uncleanSrcBasePath (toString p));
```
winterqt, line 91: Am I correct in saying that this was added in https://github.com/ipetkov/crane/commit/4f2b1c43c94e4efc72be52ca1d7bacc07fea9ab1 but wasn't actually used (I'm assuming it was just leftover from an iteration of that commit)?

ipetkov: Yep I think I was iterating on whether to make the filter on the next line as its own variable, but it looks like it was never used (since Nix is lazy it just ignores unused input :shrug: )

lib/mkDummySrc.nix
```diff
@@ -78,17 +78,25 @@ let
...
+  # when we try to crawl the source Nix will evaluate the filter(s) fully resulting in a store path
+  # whose prefix won't match the paths we observe when we try to clean the source a bit further down
+  # (Nix optimizes multiple filters by running them all once against the original source).
+  # https://github.com/ipetkov/crane/issues/46
+  origSrc =
+    if src ? _isLibCleanSourceWith
+    then src.origSrc
+    else src;
+
+  uncleanSrcBasePath = (toString origSrc) + "/";
```
winterqt, line 92: I assume this is done to forcefully evaluate the source derivation? Wouldn't this already be evaluated in the call to `findCargoFiles`?

winterqt (approved):
Tested and works, thanks!

lib/mkDummySrc.nix
ipetkov: I don't have a good enough intuitive understanding of how Nix treats paths and strings (and coercing between the two) to explain why this is needed... I did try taking it out and a bunch of tests failed so it's load bearing at some capacity :sweat_smile:

PR #49, opened 2022-07-23
Update default flags passed to cargo

ipetkov:
* The `--workspace` flag is no longer set for cargo invocations by default. Please set `cargoExtraArgs = "--workspace";` to bring this behavior back
* The `$CARGO_PROFILE` environment variable can be used to specify exactly which cargo profile is used for the invocation. By default the `release` profile will be used. Set `CARGO_PROFILE = "something else";` in the derivation to change it, or set it to `null` to omit specifying a profile altogether
* All cargo invocations are now automatically logged

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Changed
+* **Breaking**: the `--workspace` flag is no longer set for all cargo commands
+  by default. The previous behavior can be recovered by setting `cargoExtraArgs
+  = "--workspace";` in any derivation.
+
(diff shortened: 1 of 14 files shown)


dpc:
Noted. BTW. Is there any way I can have my flakes pointing at the specific version of `crane`?

ipetkov:
@dpc you can instruct a flake to use any arbitrary revision (could be a commit, or a tag, or a branch) like so:

```nix
{
  inputs.crane.url = "github:ipetkov/crane/THE_REVISION";
}
```

Also worth noting that if you are committing the `flake.lock` file (which you should be) the inputs will not change until `nix flake update` is run. Though specifying an exact revision in the `flake.nix` file will avoid accidentally updating things. Hope this helps!

PR #53, opened 2022-07-25
Add support for cargo audit

ngerstle-cognite:
This adds support for cargo-audit as a possible check when using crane. Cargo-audit uses a local clone of a repository of security advisories- this must be provided, but doing so means that crane doesn't need to be updated with repository updates.

Example usage:
```
{
  inputs = {
  ...
    crane = {
      url = "github:ipetkov/crane";
      inputs.nixpkgs.follows = "nixpkgs";
    };
    advisory-db = {
      url = "github:rustsec/advisory-db";
      flake = false;
    };
   ...
  };  
  outputs = {
    self,
    nixpkgs,
    crane,
    advisory-db,
    ...
  }:
    flake-utils.lib.eachDefaultSystem (system: let
      pkgs = import nixpkgs {inherit system overlays;};
      craneLib = (crane.mkLib pkgs).overrideToolchain rust;
      src = ./.;
      ...

    in rec {
      ...
      
      checks =
        {
          # Check formatting
          app-fmt = craneLib.cargoFmt {
            inherit src;
          };
         app-audit = craneLib.cargoAudit {
            inherit src;
            inherit advisory-db;
          }; 
          ...
        }
        ...
    });
}
```

I'm happy to consider alternatives, if there's something that seems to fit better/is more idiomatic.

[EDIT: updated example]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## Unreleased
 
 ### Added
+* Added `cargoAudit` for running [cargo-audit](https://crates.io/crates/cargo-audit)
+  with a provided advisory database instance.
+
+### Added
 * Added `cargoNextest` for running tests via [cargo-nextest](https://nexte.st/)
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## Unreleased
 
 ### Added
+* Added `cargoNextest` for running tests via [cargo-nextest](https://nexte.st/)
 * Added `cargoAudit` for running [cargo-audit](https://crates.io/crates/cargo-audit)
   with a provided advisory database instance.
-
-### Added
-* Added `cargoNextest` for running tests via [cargo-nextest](https://nexte.st/)
 
 ### Changed
 * **Breaking**: the `--workspace` flag is no longer set for all cargo commands
--- a/docs/API.md
+++ b/docs/API.md
@@ -184,6 +184,8 @@ workspace.
 
 Except where noted below, all derivation attributes are delegated to
 `cargoBuild`, and can be used to influence its behavior.
+* `cargoVendorDir` will be instantiated via `vendorCargoDeps` if not specified
+  - `cargoAuditExtraArgs` will be removed before delegating to `vendorCargoDeps`
 * `advisory-db` will be passed to the call cargo-audit as the advisory database
   (a git repo).
 * `cargoBuildCommand` will be set to run `cargo audit -n -d ${advisory-db}` in
@@ -191,6 +193,7 @@ Except where noted below, all derivation attributes are delegated to
 * `cargoExtraArgs` will have `cargoAuditExtraArgs` appended to it
   - Default value: `""`
 * `doCheck` is disabled
+* `doInstallCargoArtifacts` is disabled
 * `pnameSuffix` will be set to `"-audit"`
 
 #### Optional attributes
--- a/docs/API.md
+++ b/docs/API.md
@@ -176,6 +176,40 @@ log.
 The following hooks are automatically added as native build inputs:
 * `installFromCargoBuildLogHook`
 
+### `lib.cargoAudit`
+`cargoAudi :: set -> drv`
+
+Create a derivation which will run a `cargo audit` invocation in a cargo
+workspace.
+
+Except where noted below, all derivation attributes are delegated to
+`cargoBuild`, and can be used to influence its behavior.
+* `cargoBuildCommand` will be set to run `cargo audit -n -d ${advisory-db}` in
+  the workspace.
+* `cargoExtraArgs` will have `cargoAuditExtraArgs` appended to it
+  - Default value: `""`
+* `doCheck` is disabled
+* `pnameSuffix` will be set to `"-audit"`
+* `advisory-db` will be passed to the call cargo-audit as the advisory database
+  (a git repo).
+
+#### Optional attributes
+* `cargoAuditExtraArgs`: additional flags to be passed in the cargo-audit invocation
+  - Default value: `""`
+* `cargoExtraArgs`: additional flags to be passed in the cargo invocation
+  - Default value: `""`
+
+#### Native build dependencies
+The `cargo-audit` package is automatically appended as a native build input to any
+other `nativeBuildInputs` specified by the caller.
+
+#### Remove attributes
+The following attributes will be removed before being lowered to
+`cargoBuild`. If you absolutely need these attributes present as
+environment variables during the build, you can bring them back via
+`.overrideAttrs`.
+* `cargoAuditExtraArgs`
+
 ### `lib.cargoBuild`
 
 `cargoBuild :: set -> drv`
--- a/docs/API.md
+++ b/docs/API.md
@@ -177,21 +177,21 @@ The following hooks are automatically added as native build inputs:
 * `installFromCargoBuildLogHook`
 
 ### `lib.cargoAudit`
-`cargoAudi :: set -> drv`
+`cargoAudit :: set -> drv`
 
 Create a derivation which will run a `cargo audit` invocation in a cargo
 workspace.
 
 Except where noted below, all derivation attributes are delegated to
 `cargoBuild`, and can be used to influence its behavior.
+* `advisory-db` will be passed to the call cargo-audit as the advisory database
+  (a git repo).
 * `cargoBuildCommand` will be set to run `cargo audit -n -d ${advisory-db}` in
   the workspace.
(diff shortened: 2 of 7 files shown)


flake.lock
```diff
@@ -33,11 +33,11 @@
...
     "nixpkgs": {
       "locked": {
-        "lastModified": 1658311025,
-        "narHash": "sha256-GqagY5YmaZB3YaO41kKcQhe5RcpS83wnsW8iCu5Znqo=",
+        "lastModified": 1658644204,
+        "narHash": "sha256-MWyfCH9K3eVTXJUxBi67OQSAh9jJAnvWklM6qm4j8w8=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "cd8d1784506a7c7eb0796772b73437e0b82fad57",
+        "rev": "2f0c3be57c348f4cfd8820f2d189e29a685d9c41",
```
ipetkov, line 40: Just curious, was there a specific change in nixpkgs that we need here?

lib/cargoAudit.nix
```diff
@@ -0,0 +1,23 @@
...
+   , cargoExtraArgs ? ""
+   , advisory-db
+   , ...
+   } @ origArgs:
+let
+  args = builtins.removeAttrs origArgs [ "cargoAuditExtraArgs" ];
+in
+cargoBuild (args
+  // {
+  inherit cargoArtifacts;
```
ipetkov, line 15: `cargo-audit` should just need the `Cargo.toml` and advisory database, right? I think it should be possible to run the check without building all the dependencies first:

- we should omit `cargoArtifacts` from the function arguments
- we should overwrite `cargoArtifacts = null` so `cargoBuild` does not try to build the dependencies internally

ipetkov (commented):
Thanks for the PR @ngerstle-cognite, this is pretty much exactly as I would have imagined implementing it!

Would you be interested in adding a test case for this? We could also update the example template with using an audit job...

Happy to help finish up anything if you get stuck!

flake.lock
ngerstle-cognite: Just a routine `nix flake update` by habit, but will revert, I'm not aware of a needed change.

ngerstle-cognite:
> Thanks for the PR @ngerstle-cognite, this is pretty much exactly as I would have imagined implementing it!

Happy to help!
 
> Would you be interested in adding a test case for this? We could also update the example template with using an audit job...

I've tried to copy the test I saw for tarpaulin- making sure `cargoAudit` is run as part of `nix flake check`- is that sufficient?
I've also added it to the `quick-start` example, but only that example to start.

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Added
+* Added `cargoAudit` for running [cargo-audit](https://crates.io/crates/cargo-audit)
+  with a provided advisory database instance.
+
 ### Added
 * Added `cargoNextest` for running tests via [cargo-nextest](https://nexte.st/)
```
ipetkov, line 14:
```suggestion
### Added
* Added `cargoNextest` for running tests via [cargo-nextest](https://nexte.st/)
* Added `cargoAudit` for running [cargo-audit](https://crates.io/crates/cargo-audit)
  with a provided advisory database instance.
```

docs/API.md
```diff
@@ -176,6 +176,40 @@ log.
 The following hooks are automatically added as native build inputs:
 * `installFromCargoBuildLogHook`
 
+### `lib.cargoAudit`
+`cargoAudi :: set -> drv`
```
ipetkov, line 180:
```suggestion
`cargoAudid :: set -> drv`
```

docs/API.md
```diff
@@ -176,6 +176,40 @@ log.
...
+Except where noted below, all derivation attributes are delegated to
+`cargoBuild`, and can be used to influence its behavior.
+* `cargoBuildCommand` will be set to run `cargo audit -n -d ${advisory-db}` in
+  the workspace.
+* `cargoExtraArgs` will have `cargoAuditExtraArgs` appended to it
+  - Default value: `""`
+* `doCheck` is disabled
+* `pnameSuffix` will be set to `"-audit"`
+* `advisory-db` will be passed to the call cargo-audit as the advisory database
+  (a git repo).
```
ipetkov, line 194: (Just alpha-sorting these)

```suggestion
* `advisory-db` will be passed to the call cargo-audit as the advisory database
  (a git repo).
* `cargoBuildCommand` will be set to run `cargo audit -n -d ${advisory-db}` in
  the workspace.
* `cargoExtraArgs` will have `cargoAuditExtraArgs` appended to it
  - Default value: `""`
* `doCheck` is disabled
* `pnameSuffix` will be set to `"-audit"`
```

lib/default.nix
```diff
@@ -16,6 +16,7 @@ in
   buildPackage = callPackage ./buildPackage.nix { };
   cargoBuild = callPackage ./cargoBuild.nix { };
   cargoClippy = callPackage ./cargoClippy.nix { };
+  cargoAudit = callPackage ./cargoAudit.nix { };
```
ipetkov, line 19:
```suggestion
  cargoAudit = callPackage ./cargoAudit.nix { };
  cargoBuild = callPackage ./cargoBuild.nix { };
  cargoClippy = callPackage ./cargoClippy.nix { };
```

lib/cargoAudit.nix
ipetkov: `cargoArtifacts` represents _building the crate and its dependencies_ (think of it as the `./target/` directory)

`cargoVendorDir` represents making all crate dependency sources available in the builder standbox.

I think as long as we keep `cargoVendorDir` (which we are, or rather it will default to vendoring things from `Cargo.lock`) I think the scan should work, hence why I'm curious to skip the `buildDepsOnly` call as cargo-audit should be able to run without compiling anything.

ipetkov:
> I've tried to copy the test I saw for tarpaulin- making sure `cargoAudit` is run as part of `nix flake check`- is that sufficient?
> I've also added it to the `quick-start` example, but only that example to start.

Yep this is great, thanks!

docs/API.md
ngerstle-cognite:
```suggestion
`cargoAudit :: set -> drv`
```

ngerstle-cognite:
>I think as long as we keep cargoVendorDir (which we are, or rather it will default to vendoring things from Cargo.lock) I think the scan should work, hence why I'm curious to skip the buildDepsOnly call as cargo-audit should be able to run without compiling anything.

I agree that it should be sufficient, but I haven't yet had success when testing locally (setting `cargoArtifacts = null` and passing in `cargoVendorDir`). I'll keep trying.

ipetkov:
I can also take a closer look myself this weekend, maybe there is something I'm missing (haven't had a chance to try the PR locally yet 😅 )

ipetkov:
@ngerstle-cognite I've figured out the issue: we need to set an extra flag in the derivation:

```
doInstallCargoArtifacts = false; # Cargo audit does not generate artifacts
```

By default `cargoBuild` will try to install the artifacts to the derivation's output and since `cargo audit` doesn't produce any the build would fail

ngerstle-cognite:
> ```
> doInstallCargoArtifacts = false; # Cargo audit does not generate artifacts
> ```

This was exactly the issue! I've update the library, examples, check, and docs- I hope everything lines up?

lib/cargoAudit.nix
```diff
@@ -1,22 +1,24 @@
...
    } @ origArgs:
 let
   args = builtins.removeAttrs origArgs [ "cargoAuditExtraArgs" ];
 in
-cargoBuild (args
-  // {
-  inherit cargoArtifacts;
+cargoBuild (args // {
+  cargoVendorDir = args.cargoVendorDir or (vendorCargoDeps args);
+
```
ipetkov, line 15: `cargoBuild` already does this for us automatically so we can omit this :)
```suggestion
```

lib/cargoAudit.nix
```diff
@@ -1,22 +1,24 @@
 { cargoBuild
 , cargo-audit
+, vendorCargoDeps
```
ipetkov, line 3: See below

```suggestion
```

docs/API.md
```diff
@@ -184,13 +184,16 @@ workspace.
 
 Except where noted below, all derivation attributes are delegated to
 `cargoBuild`, and can be used to influence its behavior.
+* `cargoVendorDir` will be instantiated via `vendorCargoDeps` if not specified
+  - `cargoAuditExtraArgs` will be removed before delegating to `vendorCargoDeps`
```
ipetkov, line 188: This is fine to omit
```suggestion
```

ipetkov (approved):
Looking good, thanks again for this change @ngerstle-cognite!

Last few clean up comments and then I'm happy to merge

lib/cargoAudit.nix
ngerstle-cognite: good point- it's gone :smile:

PR #56, opened 2022-07-30
cargoAudit: filter source to only include Cargo.lock files

ipetkov:
## Motivation
Seems like cargo-audit only needs a `Cargo.lock` file and an advisory
  database to run, so we can filter the inputs down even further to
  avoid rebuilds and file copying into the Nix sandbox

## Checklist
- [x] updated `CHANGELOG.md`

cc @ngerstle-cognite

### DIFF
--- a/lib/cargoAudit.nix
+++ b/lib/cargoAudit.nix
@@ -1,22 +1,39 @@
-{ cargoBuild
-, cargo-audit
-,
-}: { cargoAuditExtraArgs ? ""
-   , cargoExtraArgs ? ""
-   , advisory-db
-   , ...
-   } @ origArgs:
+{ cargo-audit
+, cargoBuild
+, lib
+}:
+
+{ advisory-db
+, cargoAuditExtraArgs ? ""
+, cargoExtraArgs ? ""
+, src
+, ...
+}@origArgs:
 let
   args = builtins.removeAttrs origArgs [ "cargoAuditExtraArgs" ];
 in
 cargoBuild (args // {
-  cargoArtifacts = null;
   cargoBuildCommand = "cargo audit -n -d ${advisory-db}";
   cargoExtraArgs = "${cargoExtraArgs} ${cargoAuditExtraArgs}";
 
+  src = lib.cleanSourceWith {
+    inherit src;
+    # Keep all Cargo.lock files in the source in case the caller wants to
+    # pass a flag to audit a specific one.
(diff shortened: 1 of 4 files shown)


lib/cargoAudit.nix
```diff
@@ -1,22 +1,39 @@
...
+  cargoArtifacts = null; # Don't need artifacts, just Cargo.lock
+  cargoVendorDir = null; # Don't need dependencies either
   doCheck = false; # We don't need to run tests to benefit from `cargo audit`
   doInstallCargoArtifacts = false; # We don't expect to/need to install artifacts
   pnameSuffix = "-audit";
 
+  # Avoid trying to introspect the Cargo.toml file as it won't exist in the
+  # filtered source (it also might not exist in the original source either).
+  # So just use some placeholders here in case the caller did not set them.
+  pname = args.pname or "cargo";
```
ngerstle-cognite, line 35: perhaps something a tad more descriptive?
```suggestion
  pname = args.pname or "cargo-audit-results";
```

lib/cargoAudit.nix
```diff
@@ -1,22 +1,39 @@
...
-   } @ origArgs:
+{ cargo-audit
+, cargoBuild
+, lib
+}:
+
+{ advisory-db
+, cargoAuditExtraArgs ? ""
+, cargoExtraArgs ? ""
+, src
```
ngerstle-cognite, line 9: I'd be hesitant to change the order of arguments, but it should be safe enough at present.

ngerstle-cognite (approved):
Looks fine to me, and tested out locally :heavy_check_mark:

lib/cargoAudit.nix
ipetkov: Since `pnameSuffix = "-audit"` this change would make the final name `cargo-audit-results-audit`.

Having `cargo` as the placeholder is a bit weird but at least it would give us `cargo-audit` as the derivation name which I thought was a bit cleaner. Though now that I think about it it may be confusing with the derivation to build `cargo-audit` itself so I may try to rename this 😄

lib/cargoAudit.nix
ipetkov: Since we're destructuring the attrset passed in I don't see any issues (Nix doesn't have positional attrset members as far as I know)

lib/cargoAudit.nix
ipetkov: Renamed to `crate` instead (for `crate-audit`) which should be a bit less misleading!

Issue #111, opened 2022-09-21
Some way to have `buildDepsOnly` preserve extra files


dpc:
We need to use some wrappers around `linker`/`ar` binaries in `.cargo/config.toml`:

```toml
[target.armv7-linux-androideabi]
linker = "./.cargo/ld.armv7-linux-androideabi"
ar = "./.cargo/ar.armv7-linux-androideabi"

[target.aarch64-linux-android]
linker = "./.cargo/ld.aarch64-linux-android"
ar = "./.cargo/ar.aarch64-linux-android"

[target.x86_64-linux-android]
linker = "./.cargo/ld.x86_64-linux-android"
ar = "./.cargo/ar.x86_64-linux-android"

[target.i686-linux-android]
linker = "./.cargo/ld.i686-linux-android"
ar = "./.cargo/ar.i686-linux-android"
```

These are some custom scripts that use env variables to fix some Android toolchain issues.

Unfortunately, I discovered that `mkDummySrc` is not friendly to these, and all of them are stripped from the building src, so building project dependencies fails.

Some escape hatch, and maybe by default preserving whole `./cargo/` directory would be nice.

ipetkov:
Thank you for the suggestion _and_ the PR @dpc !

I think this is a good use case to support and I'm open to adding an "escape" hatch for anyone who wants to customize the dummy sources that are being used.

That said, for the time being I'd like to keep the default behavior on the conservative side. It's not entirely clear to me what the ecosystem norms are around the contents of `.cargo` (besides `config.toml`). Some projects may put linker scripts there (like in your example!) but other times folks will put stuff like `audit.toml` which IMO does not make sense to rebuild the entire project if they change a few configuration lines there.

If there are particular patterns which consistently show up where it would be a good idea to keep extra files in the `.cargo` directory I'd be open to make them part of the default behavior in the future! For now I've opened #116 to allow for arbitrary customization of the dummy sources if you'd like to try it out!

dpc:
> It's not entirely clear to me what the ecosystem norms are around the contents of `.cargo` (besides `config.toml`)

I don't think they are any. 99.9% `.cargo/` is either empty or contain only one config file.

When people tweak things there, they usually refer to stuff either via absolute paths (in a non-portable, non-nix ways), or to stuff that is expected to be in `PATH` (it just occurred to me I might have pursued this approach instead, though I like it slightly less).

[...]

PR #114, opened 2022-09-22
Copy all files from `./config`

dpc:
Typically `./.cargo/` would contain only `config.toml`, but if there are any other files in there, it's highly likely they are wrappers used by the config itself that should be included in the dependency-only build anyway.

Re: https://github.com/ipetkov/crane/issues/111

## Checklist

- [x] updated `CHANGELOG.md`


dpc:
#116 works even better, methinks.

PR #116, opened 2022-09-23
mkDummySrc: allow running a custom script after dummification is done

### LINKED ISSUE #111: Some way to have `buildDepsOnly` preserve extra files

ipetkov:
## Motivation
* Allow customizing `buildDepsOnly` by specifying an already dummified source
* Allow customizing `mkDummySrc` by accepting an additional script to run

Fixes #111

## Checklist

- [x] updated `CHANGELOG.md`


ipetkov:
This is still missing a final regression test, I'll add one a bit later but before merging

dpc:
I think that will work for me.

One thing that would be nice is a one-liner example of such custom script (e.g. copying some source file) somewhere. It's unclear to me what should be used as a "source" (just cwd?) and "destination" in this shell script to copy a file.

docs/API.md
```diff
@@ -800,6 +804,10 @@ build caches. More specifically:
 #### Optional attributes
 * `cargoLock`: a path to a Cargo.lock file
   - Default value: `src + /Cargo.lock`
+* `extraScript`: additional shell script which will be run inside the builder
+  verbatim. Useful for customizing what the dummy sources include by running any
+  arbitrary commands.
```
dpc, line 809: Copy stuff from `.` into `$out/`, right?

dpc: Seems like `cwd` is `/build` and is empty at this point...

dpc: Yeah, the original src is in `${cleanSrc}` which this script does not have available around methinks.

ipetkov: Yeah this is a good thing to document as well, the builder is running from inside of `runCommand` which does not unpack the source. If you did, then you'd depend on the entire source and any change would rebuild everything

You can _carefully_ copy specific files by doing something like `"cp ${./orig_src/path/to/file} $out/foo"` which will carefully scope the dependencies _just_ to `./orig_src/path/to/file`

dpc: But ... how am I supposed to get `orig_src`? That seems like something that was available on Nix level, and no longer available in that bash hook... ?

dpc: Oh... I think I get it. I can just use `${./.}/`.

Please bare with me. I'm just a bit slow. :D

dpc: Is it normal and expected that globs don't work in that function? I have to copy bunch of files from a directory (but not the directory itself) and it turns surprisingly difficult without `*` :D

ipetkov: Yep the ${./whatever} is probably the approach you'd need, but to avoid depending on too much I'd do something like

```nix
{
  extraScript = ''
    cp -r ${./.cargo} $out
  '';
}
```

If you need a ton of other files maybe it would be easier to run the original source through `cleanSourceWith` and then copy over the entire result (which would still avoid other change invalidating the cache)

I need to think of a good way to document this with a few examples

dpc: `cp -r ${./.cargo} $out` creates `a30hlsvk69fpcy27cqdy5rwwjalm4bx1-.cargo` in `$out`. But `cp -r ${./.}/.cargo $out` works so all good. Thanks! Much better than my `find -exec cp ...`

dpc: Anyway - it all did work for me.

dpc: One thing I want to point out is that it seems like the files copied by the scripts by default seem to loose their `+x`.

ipetkov: > But `cp -r ${./.}/.cargo $out` works so all good.

I think doing it this way will cause a rebuild when anything changes in the entire `./.` directory.

Ah yeah I forgot that `./.cargo` will get turned into `/nix/store/...-.cargo`, I think the way to force `cp` to do the right thing is to invoke it as:

```nix
''
  cp -r ${./.cargo} -T $out/.cargo
''
```

dpc: `-T`? TIL. Will try it out.

dpc: Worked.

ipetkov: Nice, I'll add a version of this as a doc example

ipetkov: Example added!

PR #126, opened 2022-10-07
mkDummySrc: support freestanding targets

rvolosatovs:
Currently, when building for targets that do not support `std`, like `x86_64-unknown-none`, `cargo check` on the costructed dummy source fails with:

```
 > error[E0463]: can't find crate for `std`
       >   |
       >   = note: the `x86_64-unknown-none` target may not support the standard library
       >   = note: `std` is required by `sallyport` because it does not declare `#![no_std]`
       >   = help: consider building the standard library from source with `cargo build -Zbuild-std`
       >
       > For more information about this error, try `rustc --explain E0463`.
```

To prevent this issue from happening, set `#![no_std]` for freestanding targets.

Signed-off-by: Roman Volosatovs <rvolosatovs@riseup.net>

## Checklist

- [x] updated `CHANGELOG.md`


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+* Freestanding targets are now supported
```
ipetkov, line 9:
```suggestion
### Fixed
* Freestanding (`#![no_std]`) targets are now supported
```

ipetkov (commented):
Thanks for the contribution @rvolosatovs ! Would you be interested in adding a test case to avoid future regressions?

Happy to help if you have any questions!

lib/mkDummySrc.nix
```diff
@@ -122,6 +122,13 @@ let
 
   dummyrs = writeText "dummy.rs" ''
     #![allow(dead_code)]
+    #![cfg_attr(target_os = "none", no_std)]
+
+    #[cfg_attr(target_os = "none", panic_handler)]
+    fn panic(_info: &core::panic::PanicInfo<'_>) -> ! {
```
newAM, line 128: Probably a good idea to make the namespace explicit.

```suggestion
    fn panic(_info: &::core::panic::PanicInfo<'_>) -> ! {
```

newAM (approved):
Tested with a `thumbv7em-none-eabi` target, this will close #127, thanks for opening this @rvolosatovs!

rvolosatovs:
> Would you be interested in adding a test case to avoid future regressions?

I'd love to, if fact I looked into doing that while working on this, but was not clear how to proceed, so I did not.
I suppose the way to do it would be to add a freestanding target to the CI toolchain to ensure it's picked up with `--all-targets`?

newAM:
> I'd love to, if fact I looked into doing that while working on this, but was not clear how to proceed, so I did not. I suppose the way to do it would be to add a freestanding target to the CI toolchain to ensure it's picked up with `--all-targets`?

I forked your branch, rebased it onto the latest master, and added a commit with a cortex-m `#![no_std]` example: https://github.com/newAM/crane/tree/fix/no_std

The examples get tested in CI as far as I can tell.

Feel free to take that commit, or remix it however you want :smile:

rvolosatovs:
@ipetkov apologies for the delay, but I have just updated the PR, PTAL.

Note that the issue is not in `#![no_std]` *crates*, it's in compiling for freestanding *targets*. Freestanding targets lack a standard library and therefore cannot be compiled *for*:
1.) when `no_std` is not set
2.) when the bare-minimum required runtime handlers are are not defined

rvolosatovs:
Hmm.. I failed to reproduce the error in the check, let me do another iteration

ipetkov:
This looks great, thanks again @rvolosatovs!

I've pushed some very minor fixes which weren't even worth mentioning here, will merge as soon as CI passes!

PR #158, opened 2022-11-09
downloadCargoPackageFromGit: Ignore crates with duplicate package IDs

zhaofengli:
## Motivation

The Cargo git repo contains numerous "case-0.1.0" Cargo.toml files in the testsuite (ex: [1](https://github.com/rust-lang/cargo/blob/5ccea519d43d09c0edf4b5dffc32e068fbf46296/tests/testsuite/init/simple_git/out/Cargo.toml) [2](https://github.com/rust-lang/cargo/blob/5ccea519d43d09c0edf4b5dffc32e068fbf46296/tests/testsuite/init/simple_bin/out/Cargo.toml) [3](https://github.com/rust-lang/cargo/blob/5ccea519d43d09c0edf4b5dffc32e068fbf46296/tests/testsuite/init/simple_hg/out/Cargo.toml)). This causes permission denials when we try to overwrite the previous ones:

```nix
let
  crane = builtins.getFlake "git+https://github.com/ipetkov/crane?rev=b70e77d2e2d480a3a0bce3ecd2d981679588b23f";
  drv = crane.lib.x86_64-linux.downloadCargoPackageFromGit {
    git = "git+https://github.com/rust-lang/cargo";
    rev = "17f8088d6eafd82349630a8de8cc6efe03abf5fb";
  };
in drv
```

```console
$ nix-build repro.nix
this derivation will be built:
  /nix/store/azhwsfib08h7q136qb2n8g96448apxz8-cargo-git.drv
building '/nix/store/azhwsfib08h7q136qb2n8g96448apxz8-cargo-git.drv'...
cp: cannot create regular file '/nix/store/1a2iy68fs58v8qvqq6scdmhpbkqbhmwi-cargo-git/case-0.1.0/out/.ignore': Permission denied
cp: cannot create regular file '/nix/store/1a2iy68fs58v8qvqq6scdmhpbkqbhmwi-cargo-git/case-0.1.0/out/Cargo.toml': Permission denied
cp: cannot create regular file '/nix/store/1a2iy68fs58v8qvqq6scdmhpbkqbhmwi-cargo-git/case-0.1.0/out/src/lib.rs': Permission denied
```

Even without permission issues, the files would be mixed together. Let's ignore the duplicates like [what Cargo does](https://github.com/rust-lang/cargo/blob/5ccea519d43d09c0edf4b5dffc32e068fbf46296/src/cargo/ops/cargo_read_manifest.rs#L186-L195).

## Checklist

- [ ] updated `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
     `installCargoArtifactsMode` is defined, in which case they will be honored)
 * `registryFromGitIndex` now uses shallow checkouts for better performance
 
+### Fixed
+* Unpacking a git repository now ignores duplicate crates to match cargo's
+  behavior
+
 ## [0.9.0] - 2022-10-29
 
 ### Changed
--- a/extra-tests/fetch-cargo-git/flake.nix
+++ b/extra-tests/fetch-cargo-git/flake.nix
@@ -10,7 +10,7 @@
       pkgs = nixpkgs.legacyPackages.${system};
     in {
       packages.cargo-git = crane.lib.x86_64-linux.downloadCargoPackageFromGit {
-        git = "git+https://github.com/rust-lang/cargo";
+        git = "https://github.com/rust-lang/cargo";
         rev = "17f8088d6eafd82349630a8de8cc6efe03abf5fb";
       };
     });
--- a/extra-tests/fetch-cargo-git/flake.nix
+++ b/extra-tests/fetch-cargo-git/flake.nix
@@ -1,0 +1,17 @@
+{
+  inputs = {
+    crane.url = "github:ipetkov/crane";
+    nixpkgs.follows = "crane/nixpkgs";
+    flake-utils.follows = "crane/flake-utils";
+  };
+
+  outputs = { flake-utils, crane, nixpkgs, ... }: flake-utils.lib.eachDefaultSystem
+    (system: let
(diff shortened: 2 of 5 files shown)


zhaofengli:
From the [source](https://github.com/rust-lang/cargo/blob/5ccea519d43d09c0edf4b5dffc32e068fbf46296/src/cargo/ops/cargo_read_manifest.rs#L110 it appears that Cargo recursively traverses the subdirectories, accumulating the packages it encounters in a HashMap. When reading the manifest, it ignores packages already in the HashMap (i.e., previously encountered). For the walk it uses `fs::read_dir` whose ordering is indeterminate so I guess our `find . -name Cargo.toml` is close enough.

ipetkov:
I kinda hoped cargo would guarantee the walk format (e.g. depth first) but I suppose we can revisit this if things break...

ipetkov:
Also, is it possible to add a test case for this so we don't regress? It's fine to pin to a specific commit (e.g. to cargo's test suite)

zhaofengli:
> I kinda hoped cargo would guarantee the walk format (e.g. depth first) but I suppose we can revisit this if things break...

It's depth-first but the ordering of the paths at the same level/directory isn't determinate.

ipetkov:
> It's depth-first but the ordering of the paths at the same level/directory isn't deterministic.

I suppose "having duplicate crate names" is already a bit of a degenerate case (i.e. normal projects shouldn't have meaningful crates which collide) so we probably don't have to worry too much about it for the time being.

Will merge this in as soon as CI passes!

PR #162, opened 2022-11-10
Allow passing curlOptsList per registry, closes #159

fasterthanlime:
## Motivation

So we don't want to merge this as-is, but this is just my POC for #159.

Most of the diff is noise because I have "nixfmt on save", I'll try and redo it without it maybe? but I'm not sure what the formatting rules for this repo are either so 🤷

Essentially:

  * `registryFromDownloadUrl` now takes an optional `curlOptsList` (defaults to `[]`)
  * the entries in `crateRegistries` are no longer strings, but a record with keys `downloadUrl` (what the value was before) and `curlOptsList`
  * `urlForCargoPackage` no longer returns a string, but a record with keys `url` (what the value was before) and `curlOptsList`
  * `downloadCargoPackage` has been changed to forward `curlOptsList` to `fetchurl` and use `(urlForCargoPackage args).url`

Apart from the formatting noise, open questions are:

  * should `urlForCargoPackage` be renamed? it no longer returns _just_ the URL
  * what happens if the user's `fetchurl` doesn't support `curlOptsList`?

Anyway this works for me for now, I'm happy to address feedback so this can land, or to wait till you make your own PR @ipetkov.

## Checklist

- [ ] added tests to verify new behavior
- [ ] added an example template or updated an existing one
- [ ] updated `docs/API.md` with changes
- [ ] updated `CHANGELOG.md`

PR #168, opened 2022-11-18
Allow passing arbitrary fetchurl args per registry

### LINKED ISSUE #159: Allow configuring additional curl parameters per registry
[Shipyard](https://shipyard.rs/) is a private cargo registry hosting solution. Currently it requires [passing an authentication token via user agent string](https://docs.shipyard.rs/authentication/temporary-fix.html)

We should support a way of defining additional curl parameters (per registry) which can be passed into `fetchUrl` while downloading the crate sources

ipetkov:
## Motivation
Same as #162 but with my own updates pushed to it

Closes #159 

## Checklist
- [x] updated `CHANGELOG.md`


ipetkov:
Note: still need to update the docs and changelog before merging

PR #183, opened 2022-12-08
zstd: use as many threads as we can when compressing

### LINKED ISSUE #182: `zstd` seems to use default (1) number of threads
We've noticed that on fast machines `zstd` usage in crane appears slow and bottlenecks on saturating only one CPU, even though `zstd` is supposed to support multi-threaded compression&decompression.

ipetkov:
## Motivation
zstd compression is supposed to be deterministic (even with multiple threads) so we might as well allow it to use as many cores as are available to the derivation

Fixes #182

## Checklist

- [x] updated `CHANGELOG.md`


dpc:
Does its thing here. I see `-T20` used and it seems faster indeed.

PR #185, opened 2022-12-12
Add buildInputs to quick-start doc

mitchty:
## Motivation

Upon further testing it looks like the crate-doc quick start also needs buildInputs when you add more involved doc tests on macos. Other platforms should remain unaffected.

## Checklist

- [ ] updated `CHANGELOG.md`

### DIFF
--- a/examples/quick-start/flake.nix
+++ b/examples/quick-start/flake.nix
@@ -65,7 +65,7 @@
           };
 
           my-crate-doc = craneLib.cargoDoc {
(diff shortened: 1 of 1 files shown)


ipetkov:
Looks good, thanks for the PR @mitchty!

mitchty:
No problem I wasn't sure on my original pr and deferred to later, turns out it needed it after all on macos. Thanks for the quick turnaround!

PR #186, opened 2022-12-12
multi folder build support

dzmitry-lahoda:
## Motivation

```
./a/b/Cargo.toml - simple package without workspace
./c/d/Cargo.lock
./c/d/Cargo.toml - workspace
./c/d/e/Cargo.toml - package from workspace referencing ./a/b/Cargo.toml 
```

So we can setup toml file from any folder to allow non root folder build.

Are there other options?

## Checklist

- [ ] updated `CHANGELOG.md`

### DIFF
--- a/lib/buildDepsOnly.nix
+++ b/lib/buildDepsOnly.nix
@@ -36,7 +36,7 @@ let
   cargoCheckExtraArgs = args.cargoCheckExtraArgs or (if doCheck then "--all-targets" else "");
 
   path = args.src or throwMsg;
-  cargoToml = path + "/Cargo.toml";
+  cargoToml = args.cargoToml or (path + "/Cargo.toml");
   dummySrc = args.dummySrc or
     (if builtins.pathExists cargoToml
     then mkDummySrc args


ipetkov:
Thanks for the PR @dzmitry-lahoda! Could you elaborate a bit more on the project structure you are trying to build?

* Is there a single `flake.nix` file at the `./` root which builds all the crates together?
* Is the goal here to only build `./a/b/Cargo.toml`, or is the goal to build `./c/d/Cargo.toml` whose member has a path dependency on `./a/b/Cargo.toml`?
* Is `./a/b/Cargo.toml` a totally standalone project (i.e. does it also have a `Cargo.lock`)?

Although it is possible to allow `buildDepsOnly` to use an explicit path to a `Cargo.toml` file, it would still need to match with an appropriate `Cargo.lock` file (mixing and matching these from different workspaces _may_ work but it isn't guaranteed). Normally you'd want to pass the entire cargo workspace root to the Nix builder and then instruct cargo specifically which crates you'd want to build (e.g. by `cargo build -p mycrate`), but I'd like to understand a bit better about the concrete use-case and the troubles it is bringing!

dzmitry-lahoda:
> Although it is possible to allow buildDepsOnly to use an explicit path to a Cargo.toml file, it would still need to match with an appropriate Cargo.lock file (mixing and matching these from different workspaces may work but it isn't guaranteed). Normally you'd want to pass the entire cargo workspace root to the Nix builder and then instruct cargo specifically which crates you'd want to build (e.g. by cargo build -p mycrate), but I'd like to understand a bit better about the concrete use-case and the troubles it is bringing!

I see. May be it would be issue. For now we build contracts. So it is not so big deal.

dzmitry-lahoda:
https://github.com/ComposableFi/composable/blob/538d1c3df71083a2f38912627985683602eeaa39/cosmwasm/contracts/Cargo.lock <- contracts easy to build and small (so deps build only is not problem)

these depend on https://github.com/ComposableFi/composable/tree/538d1c3df71083a2f38912627985683602eeaa39/code/xcvm/lib/core

which is part of https://github.com/ComposableFi/composable/blob/538d1c3df71083a2f38912627985683602eeaa39/code/Cargo.lock (hard to build - we build deps - work well)

so we afraid moving `core` to separate folder not under workspace will be issue (moving to other git repo is issue too - hard to sync deps)

ipetkov:
I see, the problem is that your workspace root itself is not an ancestor of all the crates it includes (so the Nix expression cannot just scope to a sub directory), and the workspace root is not part of the flake root either. In that case I'm open to merging this!

Are you interested in adding a simple test case to avoid any future regressions here? The crate contents can be empty, it's the structure that is interesting here

dzmitry-lahoda:
1. examples of structure should be added
2. team decided to do reexport (pub use) of conflicting deps instead of split workspace (so it will bloat wasm files size and other not good things), so it is not prior for us not. so may be I will later return here when we will finally decide split (which is good IMHO)

ipetkov:
Thanks again for the PR @dzmitry-lahoda ! I've ended up addressing this in #212 which was slightly different change which simplified these checks altogether

PR #198, opened 2022-12-26
README: add FAQ about building workspace subsets

### LINKED ISSUE #197: Document how to handle Workspaces
### Discussed in https://github.com/ipetkov/crane/discussions/189

Here is some writeup of what I found out:

# Cargo Workspaces
Workspaces with multiple members are supported. The examples build everything in the workspace by default, similar as to how `cargo build` will try to build everything in the workspace by default. 

## Building Binaries in a Workspace
Binaries in a workspace can be built by adding a `--bin` flag to the `cargoExtraArgs` and setting the `pname` like this:
```nix
my-bin = craneLib.buildPackage {
    inherit cargoArtifacts src buildInputs;
    pname = "my-bin";
    cargoExtraArgs = "--bin my-bin";
};
```

ipetkov:
## Motivation
Fixes #197 

## Checklist
- [ ] updated `CHANGELOG.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -438,6 +438,25 @@ craneLib.buildPackage {
 }
 ```
 
+### How can I build only a subset of a given cargo workspace?
+
+By default, cargo will build the crate at the current directory when invoked; if
+the current directory holds a workspace, cargo will then build all crates within
+that workspace.
+
+Sometimes it can be useful to only build a subset of a given workspace (e.g.
+only specific binaries are needed, or some crates cannot be built for certain
+platforms, etc.), and cargo [can be instructed to do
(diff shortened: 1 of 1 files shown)


DrRuhe:
Could you add a note stating that when using the `--bin` flag its required to also add the pname so that the resulting detivation is configured to correctly run? Theoretically it would be possible to extract the pname from `--bin foo` usage, but that seems a bit hacky to me (although the pname thing is also not particularly obvious).

ipetkov:
Ah yes will update, thanks!

ipetkov:
Fixed in 67b1799c335f6aeff44e90c5ee463cbd93287bb1!

Issue #209, opened 2023-01-04
Building a package or workspace in a subdirectory


paulyoung:
I have a directory structure that looks like this:

```
.
├── backend
│   ├── Cargo.lock
│   ├── Cargo.toml
...
├── frontend
...
```

Up until now I've been able to set `src = ./backend` but now I need to access some files that live in the root.

I've gotten as far as this:

```nix
{
  src = ./.;
  cargoToml = ./backend/Cargo.toml;
  cargoLock = ./backend/Cargo.lock;
  cargoExtraArgs = "--manifest-path ./backend/Cargo.toml";
}
```

but now I seem to be stuck because `cargo manifest` doesn't get `cargoExtraArgs`: https://github.com/ipetkov/crane/blob/0017a8da6824c5f27c48cce7d614d3e590e75e35/lib/setupHooks/installFromCargoBuildLogHook.sh#L34

Is there a better way to approach this?

dpc:
Best way IMO, is the filter `src` with an explicit list of stuff to include. Example from our project: https://github.com/fedimint/fedimint/blob/65b2058e52b773f83e023a7cbb1d13988bc19197/flake.nix#L424

paulyoung:
If I understand correctly, I think you’re suggesting using a filter to make it look like the contents of the `backend` directory (in my case) are at the root. Is that correct?

If not, could you help me understand how filtering would help? The issue seems to be that `Cargo.toml` isn’t at the root.

paulyoung:
> If I understand correctly, I think you’re suggesting using a filter to make it look like the contents of the `backend` directory (in my case) are also at the root. Is that correct?

I think this approach would be problematic anyway because it would require changing relative paths that are looking for other files at the root and then that wouldn’t work in a dev shell when using `cargo` directly.

paulyoung:
I think I’m going to see if introducing `cargoManifestExtraArgs` (which would default to `cargoExtraArgs`) fixes my issue and then maybe make a PR.

dpc:
> If I understand correctly, I think you’re suggesting using a filter to make it look like the contents of the `backend` directory (in my case) are also at the root. Is that correct?

No, I was going to suggest making them a workspace, and just filtering necessary files.

paulyoung:
I already have a workspace in `./backend` and build specific crates using `-p` so not sure how that helps. I'm trying to say `src = ./.` and still have `crane` be able to find what it needs without changing any of the existing directory structure.

[...]

PR #210, opened 2023-01-05
Add `cargoMetadataExtraArgs`

paulyoung:
## Motivation

This adds `cargoMetadataExtraArgs` to support the use case described in https://github.com/ipetkov/crane/issues/209. In short, it enables building projects where `Cargo.toml` lives in a subdirectory of the root. 

This almost worked already:

```nix
{
  src = ./.;
  cargoToml = ./backend/Cargo.toml;
  cargoLock = ./backend/Cargo.lock;
  cargoExtraArgs = "--manifest-path ./backend/Cargo.toml";
}
```

but `cargoExtraArgs` wasn't being passed to `cargo manifest`.

I tried providing `cargoExtraArgs` there first, but some flags aren't supported by `cargo metadata` so it makes sense to have a separate option.

For example, I'm also providing `--package` in `cargoExtraArgs` and `cargo manifest` complains about that.

So, this pull request allows for this instead:

```nix
{
  src = ./.;
  cargoToml = ./backend/Cargo.toml;
  cargoLock = ./backend/Cargo.lock;
  cargoMetadataExtraArgs = "--manifest-path ./backend/Cargo.toml";
}
```

## Checklist

- [x] updated `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Added
+* `buildPackage` now accepts `cargoManifestExtraArgs` for passing additional
+  arguments just to the `cargo manifest` invocation
+
 ### Fixed
 * Previously compiled build scripts now maintain their executable bit when
   inherited
--- a/docs/API.md
+++ b/docs/API.md
@@ -104,6 +104,9 @@ to influence its behavior.
 * `cargoExtraArgs`: additional flags to be passed in the cargo invocation (e.g.
   enabling specific features)
   - Default value: `""`
+* `cargoMetadataExtraArgs`: additional flags to be passed in the `cargo metadata`
+  invocation (e.g. specifying `--manifest-path`)
+  - Default value: `""`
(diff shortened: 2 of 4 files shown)


paulyoung:
I should perhaps reiterate (I mention it in #209, but not here) that the motivation for the above setup is so that some files in the root directory can be accessed by the Rust build.

However; I just tried this and while it works for `cargo` in a dev shell, my code thinks the files don't exist when attempting build with `nix`.

It seems that I can't ascend to any directory above the one that contains `Cargo.toml`.

ipetkov:
Thanks for the PR @paulyoung, the approach is definitely sensible! Though with #212 landed and documented my inclination is towards closing this.

`buildPackage` is basically a glorified wrapper whose goal is to effectively call `cargoBuild` and then immediately run `installFromCargoBuildLog`. My concern is that introducing `cargoMetadataExtraArgs` as yet another lever to potentially configure would lead to more confusion than simply changing directories or having the caller provide their own custom `installPhaseCommand` (which can just as easily invoke `installFromCargoBuildLog` itself!).

Thanks again for documenting your findings and looking for ways to improve things!

paulyoung:
I completely understand. I'll take a look at #212. Thanks!

PR #234, opened 2023-02-09
ignore non-json lines in cargoBuildLog

felschr:
Fix issues with non-JSON values in `cargoBuildLog` by ignoring non-json lines.

## Motivation
In some cases using cargo with `message-format=json` prints some non-json lines: https://github.com/rust-lang/cargo/issues/8179

When that's the case crane currently fails with:
```
searching for bins/libs to install from cargo build log at cargoBuildLogGbmv.json
parse error: Invalid numeric literal at line 390, column 6
```

## Checklist

- [ ] updated `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Fixed
+* `buildPackage` is more tolerant of misbehaving proc macros which write to
+  stdout during the build
+
 ## [0.11.1] - 2023-01-21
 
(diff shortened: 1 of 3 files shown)


ipetkov:
Thanks for the contribution @felschr, overall looks good to me!

Are you interested in adding a test case for this to make sure we don't regress? It doesn't need to include a proc-macro which (incorrectly) emits data on stdout, it can be as simple as running the hook with a sample log file which includes non-json data

felschr:
Here we go, does this look good to you?

ipetkov (approved):
Looks great, thanks again!

Issue #236, opened 2023-02-10
Automatic `pname` selection for cargo workspace results in `nix run` failing with an obtuse error


benwis:
At the risk of looking silly again, I was able to get this to build fine, but I cannot get it to run with `nix-run .`
```nix
{
  description = "Build a cargo project with a custom toolchain";

  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";

    crane = {
      url = "github:ipetkov/crane";
      inputs.nixpkgs.follows = "nixpkgs";
    };

    flake-utils.url = "github:numtide/flake-utils";

    rust-overlay = {
      url = "github:oxalica/rust-overlay";
      inputs = {
        nixpkgs.follows = "nixpkgs";
        flake-utils.follows = "flake-utils";
      };
    };
    advisory-db = {
      url = "github:rustsec/advisory-db";
      flake = false;
    };
  };

  outputs = { self, nixpkgs, crane, flake-utils, advisory-db, rust-overlay, ... }:
    flake-utils.lib.eachDefaultSystem
      (system:
        let
          pkgs = import nixpkgs {
            inherit system;
            overlays = [ (import rust-overlay) ];
          };

          rustTarget = pkgs.rust-bin.selectLatestNightlyWith (toolchain: toolchain.default.override {
            extensions = [ "rust-src" "rust-analyzer" ];
            targets = [ "wasm32-unknown-unknown" ];
          });

          # NB: we don't need to overlay our custom toolchain for the *entire*
          # pkgs (which would require rebuidling anything else which uses rust).
          # Instead, we just want to update the scope that crane will use by appendings
          inherit (pkgs) lib;
          # our specific toolchain there.
          craneLib = (crane.mkLib pkgs).overrideToolchain rustTarget;
          #craneLib = crane.lib.${system};
          # Only keeps markdown files
          protoFilter = path: _type: builtins.match ".*proto$" path != null;
          sqlxFilter = path: _type: builtins.match ".*json$" path != null;
          protoOrCargo = path: type:
            (protoFilter path type) || (craneLib.filterCargoSources path type) || (sqlxFilter path type);
          # other attributes omitted
          src = lib.cleanSourceWith {
            src = ./.; # The original, unfiltered source
            filter = protoOrCargo;
          };
          #    src = craneLib.cleanCargoSource ./.;

          buildInputs = [
            # Add additional build inputs here
            pkgs.pkg-config
            pkgs.openssl
[...]

PR #237, opened 2023-02-12
crateNameFromCargoToml: warn when values can't be found

ipetkov:
## Motivation
Emit a warning whenever `crateNameFromCargoToml` fails to find an appropriate value for a package's `name` or `version` fields. This should make it a bit more obvious when this happens (i.e. so the caller can explicitly set an appropriate value and not have things like `nix run` break "inexplicably") but still allow the build to go through.

Fixes https://github.com/ipetkov/crane/issues/236

## Checklist

- [x] updated `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Changed
+* A warning will now be emitted if a derivation's `pname` or `version`
+  attributes are not set and the value cannot be loaded from the derivation's
+  root `Cargo.toml`. To resolve it consider setting `pname = "...";` or `version
+  = "...";` explicitly on the derivation.
+
 ## [0.11.2] - 2023-02-11
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


torhovland:
Is this working well with workspaces? It looks like it is only looking in the root `Cargo.toml`, where `name` and `version` are not set, and ignoring all other `Cargo.toml` files where they _are_ set.

ipetkov:
@torhovland it is intentional that only the root `Cargo.toml` is inspected, if there are multiple `Cargo.toml` files (for multiple crates) it is up to the maintainer to disambiguate which one represents the derivation.

There are two options to avoid having to sync the `Cargo.toml` and Nix values by hand:
* Set `package.name` and `package.version` in the `[workspace]` declaration of the root `Cargo.toml`
* Manually invoke `crateNameFromCargoToml` with the appropriate path and pass the values through:
```nix
let
  crateName = craneLib.crateNameFromCargoToml {
    cargoToml = ./some/subcrate/Cargo.toml;
  };
in
craneLib.buildPackage {
  inherit (crateName) pname version;

  # all other arguments go here
  src = craneLib.cleanCargoSource ./.;
}
```

PR #258, opened 2023-03-03
feat: add support for `bindeps` feature

rvolosatovs:
## Motivation

This adds support for [`bindeps` feature](https://doc.rust-lang.org/cargo/reference/unstable.html#artifact-dependencies)

I used to work around this by removing `target` and `artifact`(https://github.com/rvolosatovs/nixify/blob/559177816ea95cddf7d419c05488f5bd8021e20b/lib/rust/mkOverlay.nix#L161-L166), but that breaks in cases where e.g. you're cross-compiling for `musl` and have a bindep for `wasm32-unknown-unknown`, because `cdylib` are not supported for `musl`

You can reproduce the issue by trying to build the added example on current `master`

## Checklist

- [x] updated `CHANGELOG.md`


lib/mkDummySrc.nix
```diff
@@ -125,6 +125,7 @@ let
   dummyrs = args.dummyrs or (writeText "dummy.rs" ''
     #![allow(dead_code)]
     #![cfg_attr(any(target_os = "none", target_os = "uefi"), no_std)]
+    #![cfg_attr(any(target_os = "none", target_os = "uefi"), no_main)]
```
ipetkov, line 128: Why do we need `no_main` if we have a(n empty) main function further down?

lib/mkDummySrc.nix
```diff
@@ -195,6 +196,7 @@ let
       '' + optionalString (trimmedCargoToml ? package) ''
         # To build regular and dev dependencies (cargo build + cargo test)
         ${cpDummy parentDir "src/lib.rs"}
+        ${cpDummy parentDir "src/bin/dummy/main.rs"}
```
ipetkov, line 199: Perhaps we should name the dummy binary as `crane-dummy` so it's a _bit_ more obvious to someone where it could be coming from if they are poking through their `target` directory

```suggestion
        ${cpDummy parentDir "src/bin/crane-dummy/main.rs"}
```

examples/bindeps/foo/Cargo.lock
```diff
@@ -0,0 +1,7 @@
+# This file is automatically @generated by Cargo.
```
ipetkov, line 1: This `Cargo.lock` file is unnecessary, since `./foo` is a path dependency it will automatically get promoted to a workspace member (i.e. it's packages area already present in the parent `Cargo.lock` file)

examples/bindeps/flake.nix
```diff
@@ -0,0 +1,50 @@
...
+    rust-overlay = {
+      url = "github:oxalica/rust-overlay";
+      inputs = {
+        nixpkgs.follows = "nixpkgs";
+        flake-utils.follows = "flake-utils";
+      };
+    };
+  };
+
+  outputs = { nixpkgs, crane, flake-utils, rust-overlay, ... }:
```
ipetkov, line 23: Do you mind turning this example into a "regular" flake check test?

There isn't anything specific that consumers have to change about their `crane` invocations for bindeps (besides making sure their toolchain supports the necessary targets), so we probably don't need to make a specific template for it. Though I do want to keep it as a test!

examples/bindeps/src/main.rs
```diff
@@ -0,0 +1,8 @@
+fn main() {
+    println!(env!("CARGO_CDYLIB_FILE_FOO"));
```
ipetkov, line 2: We should include a `BIN` dependency here too to make sure it is wired up properly as well
```suggestion
    println!(env!("CARGO_CDYLIB_FILE_FOO"));
    println!(env!("CARGO_BIN_FILE_FOO"));
```

CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Added
+
+* Add `main.rs` to each "dummy" crate generated to support
```
ipetkov, line 11:
```suggestion
* Add a stubbed binary target to each "dummy" crate generated to support
```

ipetkov (commented):
Thanks for the PR @rvolosatovs!

I've never used the bindeps feature before so I wanted to leave some notes (for posterity) from my findings:

---

(Currently) depending on a crate's binary artifacts would only break for _path dependencies_, as other crates are going to get built in the deps-only derivation which shouldn't cause any problems.

One potential solution is to do what was mentioned above: clean the `Cargo.toml`s to omit any `bin`/`lib`/`artifact`/`target` attributes set for any dependency. Though this breaks down because we _do want_ to build and cache such artifacts for "external" crates. We could try to do the cleaning only for path deps, but this also isn't enough as a path dependency does not imply the source is coming from the same root (e.g. a consumer could have rewritten path deps to point to the Nix store directly).

---

A better approach is exactly what this MR does: add a dummy binary that is available for cargo to compile. We don't really care what it is called, as long as a binary is available, cargo should be happy if a `bin` dependency is declared on the crate.

One potential downside with forcing all dummy sources to contain a dummy binary is that building with `--all-targets` will result in those binaries showing up in the artifacts folder. If someone were to naively crawl cargo's `target` directory they could accidentally install some garbage. I think this downside is fairly low and can be remedied more easily if someone were to trip on it.

---

Overall the approach looks good to me! Just have some clean up comments (also there is one failing test case which needs to be fixed)

lib/mkDummySrc.nix
rvolosatovs: I am honestly not sure, the `noStd` test fails without this statement with error regarding missing "`start` lang_item"
Annotating `main` with `start` would require a nightly feature

examples/bindeps/foo/Cargo.lock
rvolosatovs: Hmm, good catch, it should not have been here.

examples/bindeps/flake.nix
rvolosatovs: That's exactly what I wanted to do initially, but it would require a custom toolchain and I have not found an example check using a custom toolchain. But I'll work on that

lib/mkDummySrc.nix
ipetkov: Hmm, that's weird, but if it works it works (we can always clean this up later, so no blocker here)

examples/bindeps/flake.nix
ipetkov: We have `rust-overlay` available as an input, actually the `noStd` test uses a custom toolchain if you want to use that as a starting point: https://github.com/ipetkov/crane/blob/105e27adb70a9890986b6d543a67761cbc1964a2/checks/default.nix#L237-L251

rvolosatovs: Right, that was the test case I added myself a while ago :laughing: 097afac

ipetkov (approved):
Looks great, thanks again!

PR #267, opened 2023-03-14
feat: add cargoLlvmCov

figsoda:
## Checklist

- [x] updated `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,9 @@ in case a crate is used as `bin` artifact dependency.
   = "...";` explicitly on the derivation.
 * A warning will now be emitted if `src` and `dummySrc` are passed to
   `buildDepsOnly` as `dummySrc` will take priority
+* cargo is now invoked with `--release` when `$CARGO_PROFILE == release` instead
+  of passing in `--profile release` to better support tools which do not
+  understand the latter
 
 ## [0.11.2] - 2023-02-11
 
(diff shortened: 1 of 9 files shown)


ipetkov:
Thanks for the PR @figsoda, this looks great, especially the updated docs and example so thank you for that polish!

I had some minor feedback which I went ahead and pushed to avoid having to go back and forth. As a quick summary:
* Changed the `cargoWithProfile` helper to prefer using `--release` over `--profile release` since `cargo-llvm-cov` doesn't appear to understand it. This also sidesteps the awkward clearing of the `CARGO_PROFILE` flag when invoking `cargoLlvmCov` (which, arguably, wasn't _quite_ correct to do since we build release artifacts by default, but clearing the flag would mean that `cargo llvm-cov` would need to rebuild everything from scratch)
* Renamed `llvmToolsLib` in the quick-start example to `craneLibLlvmTools` to make it a bit more obvious what it represents
* Removed the `fenix` input and updated the checks to use a toolchain from `rust-overlay`. I have nothing against `fenix` personally, though adding it would mean having it show up in the lockfiles of all downstream consumers. Arguably `rust-overlay` shouldn't be there either: I'd like to eventually remove any checks-related inputs out of the main lock file, but that will need to be done separately

ipetkov (approved):

figsoda:
I just didn't know how to use rust-overlay, thanks for changing that for me!

PR #271, opened 2023-03-19
feat: add support for `wit` files

rvolosatovs:
## Motivation

(sorry, clicked the "open PR" button too fast)

This PR adds support for WIT files https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md, used by WebAssembly components to generate bindings via https://github.com/bytecodealliance/wit-bindgen

Today, `.wit` files are trimmed by `crane`, which means `bindgen` macro cannot find them (it searches by paths) in the project being built, but, more importantly, in dependencies (here's an example https://github.com/wasmCloud/wasmCloud/actions/runs/4415710680/jobs/7739037159
> wasmcloud-deps> error: failed to read file "/nix/store/4f8fkwmzdh62bjr3wpl2kphq2vixdkj6-vendor-cargo-deps/225fe4a8a7631cbf30190f12e81bfc52813aac3e2e059f80292a8f336881417f/host-0.0.0/../wit"
)

An example of an "unbuildable" WIT-enabled Rust project, which is often included as a depedency is https://github.com/bytecodealliance/preview2-prototyping

Would this PR be something you'd be interested in merging @ipetkov ?

## Checklist

- [ ] updated `CHANGELOG.md`

### DIFF
--- a/lib/downloadCargoPackageFromGit.nix
+++ b/lib/downloadCargoPackageFromGit.nix
@@ -29,6 +29,11 @@ in
 runCommandLocal "cargo-git" deps ''
   mkdir -p $out
   existing_crates=()
+  for f in $(find ${repo} -type f -name '*.wit'); do
+    local path=$(realpath --relative-to=${repo} $f)
+    mkdir -p $out/$(dirname $path)
+    cp -L --copy-contents $f $out/$path;
+  done
   while read -r cargoToml; do
     local crate=$(
       cargo metadata --format-version 1 --no-deps --manifest-path "$cargoToml" |
--- a/lib/findCargoFiles.nix
+++ b/lib/findCargoFiles.nix
@@ -6,6 +6,7 @@ let
   inherit (lib)
     flatten
     groupBy
+    hasSuffix
     mapAttrs
     mapAttrsToList;
 
@@ -17,6 +18,7 @@ let
         cur = dir + "/${name}";
         isConfig = parentIsDotCargo && (name == "config" || name == "config.toml");
         isCargoToml = name == "Cargo.toml";
+        isWit = hasSuffix ".wit" name;
       in
       if type == "directory"
       then listFilesRecursive (name == ".cargo") cur
@@ -24,6 +26,8 @@ let
       then [{ path = cur; type = "cargoTomls"; }]
       else if isConfig
       then [{ path = cur; type = "cargoConfigs"; }]
+      else if isWit
+      then [{ path = cur; type = "wits"; }]
       else [ ]
     )
     (builtins.readDir dir));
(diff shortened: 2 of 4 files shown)


ipetkov:
Thanks for looking into this @rvolosatovs!

> Would this PR be something you'd be interested in merging @ipetkov ?
I definitely would like to support such (namely WASM) builds but I would really like to avoid having to special case `*.wit` files from a maintenance perspective (e.g. wouldn't want to go down the path of special casing `*.proto` files, etc.)

It looks like the crux of the issue is that the [preview2-prototyping](https://github.com/bytecodealliance/preview2-prototyping) project has all the `*.wit` files in their own directory which is NOT part of any crate name. When vendoring crates only directories with `Cargo.toml` files are extracted, so naturally that directory gets ignored. This is how `cargo vendor` behaves, though most people don't realize it because consuming crates from crates.io or as a git dependency _happens_ to copy all the files over

If by using `cargo vendor` with `preview2-prototyping` as a dependency _succeeds_ then I would consider it a `crane` bug that ought to be fixed. If, by using `cargo vendor` with with `preview2-prototyping` as a dependency _fails_ then my initial reaction would be to try to "fix" the issue upstream (e.g. maybe by symlinking the `wit` directory within every crate that uses it and avoid having the `bindgen!` macros try to look outside of the current directory?)

rvolosatovs:
> Thanks for looking into this @rvolosatovs!
> 
> > Would this PR be something you'd be interested in merging @ipetkov ?
> > I definitely would like to support such (namely WASM) builds but I would really like to avoid having to special case `*.wit` files from a maintenance perspective (e.g. wouldn't want to go down the path of special casing `*.proto` files, etc.)
> 
> It looks like the crux of the issue is that the [preview2-prototyping](https://github.com/bytecodealliance/preview2-prototyping) project has all the `*.wit` files in their own directory which is NOT part of any crate source. When vendoring crates only directories with `Cargo.toml` files are extracted, so naturally that directory gets ignored. This is how `cargo vendor` behaves, though most people don't realize it because consuming crates from crates.io or as a git dependency _happens_ to copy all the files over
> 
> If by using `cargo vendor` with `preview2-prototyping` as a dependency _succeeds_ then I would consider it a `crane` bug that ought to be fixed. If, by using `cargo vendor` with with `preview2-prototyping` as a dependency _fails_ then my initial reaction would be to try to "fix" the issue upstream (e.g. maybe by symlinking the `wit` directory within every crate that uses it and avoid having the `bindgen!` macros try to look outside of the current directory?)

You're spot on, `cargo vendor` case fails completely with a path leading outside of the crate root.

Thanks! I'll file an issue upstream!

ipetkov:
> You're spot on, `cargo vendor` case fails completely with a path leading outside of the crate root.
> 
> Thanks! I'll file an issue upstream!

Cool, thanks for confirming! I'll close this out since we've confirmed this is an upstream issue but let me know if you disagree

Issue #180, opened 2022-12-02
Add buildWorkspace


sellout:
I’m not sure if this is a good idea, but if not I’d like to hear what the best alternative is (explicit separate `buildPackage` for each lib in the workspace?)

In any case, I want roughly
```nix
  outputs = { self, crane, flake-utils, nixpkgs, ... }:
    flake-utils.lib.eachDefaultSystem (system:
      let
        craneLib = crane.lib.${system};
        pkgs = import nixpkgs {inherit system;};
      in
    {
      packages =
        {default = self.packages.${system}.oneOfTheOnesInTheWorkspace;}
        // craneLib.buildWorkspace {
          src = craneLib.cleanCargoSource ./.;
          doCheck = true;
          buildInputs = [pkgs.libiconv];
          nativeBuildInputs = [];
        };
    });
```
where `buildWorkspace` produces an attrset of packages, one per package in the workspace. And the attrs passed to `buildWorkspace` are propagated to the individual packages. Although maybe it makes sense to put propagated attrs in a single attr, so other `buildWorkspace` attrs (like, maybe an `exclude` list or something) can be provided without conflicting or being propagated.

I’m still on the steep part of the learning curve for both Nix and Rust, so I’m as interested in arguments against this as I am in seeing it done.

ipetkov:
Hi @sellout thanks for the report! Let me ask a bit more concretely: what potential advantages would you be interested in seeing from`buildWorkspace` which would be impossible with `buildPackage`? (keeping in mind that `buildPackage` can easily build all binaries in the workspace :wink: )

---
From the perspective of _organizing derivations and their outputs_ I could see the advantage of being able to select a specific "output" of the workspace (I know `crate2nix` supports this). That said, I see a couple of reasons why I wouldn't consider this direction (alone) a motivator:

* implementation costs: this would require even deeper analysis of the project structure to figure out what the outputs are (I don't know of any cargo-related tools which allow for easy/arbitrary queries of the project structure). With build-scripts I have a feeling it becomes an undecidable problem. Even if figuring out all the outputs can be solved (elegantly?) there is still the issue of knowing which targets to actually build. Some targets may depend on others, but having separate outputs is kind of moot if you have to build the whole workspace as the granularity doesn't afford you much of anything (besides maybe a derivation with one `/bin/*` output)
[...]

PR #293, opened 2023-04-07
Allow `cargoArtifacts` in `buildDepsOnly`

akiekintveld:
## Motivation
A project I am working on is structured as a single repo with multiple sub-crates, but it cannot use cargo workspaces to build them all at once (for unrelated reasons). In particular, it has a library crate and multiple application crates that have path dependencies on that library. I would like to make sure that the library is only built once by using `buildDepsOnly` on the library, and then passing the resulting derivation as `cargoArtifacts` into each application's `buildDepsOnly` to reuse the `target` directory of the library in each application. I also added a test that illustrates what I'm trying to accomplish, and it appears to exhibit the correct reuse with this change.

## Checklist
- [x] updated `CHANGELOG.md`


ipetkov:
Thank you for the PR @akiekintveld, I appreciate that you have included a test case as well as updated the documentation and changelog!

Although the approach here is not unreasonable, my hesitation in accepting this change is because I'm not sure if it meaningfully solves the mentioned use case: if any file in the common sub-directory changes it will not only rebuild the application crate itself but all of its (additional) dependencies too.

> but it cannot use cargo workspaces to build them all at once (for unrelated reasons).

It's worth mentioning that [cargo will automatically promote path dependencies as workspace members](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-members-and-exclude-fields), so I'm curious to hear more about this and whether it's a technical limitation or simply wanting to control the caches better. Currently there isn't a great story for having granular caching of individual crates within a workspace, which might be what you are running into here #180 

(It's also worth noting that anyone blocked on this can easily inject a cargo target directory via `(buildDepsOnly args).overrideAttrs (_old: { inherit cargoArtifacts; })` even though `buildDepsOnly`)

akiekintveld:
Hi! Sadly my use case isn't quite covered by #180. My project consists of a library crate and two dependent application crates, but both applications have to be built for different targets. One application is built for the host target, and the other is built for `riscv64imac-unknown-none-elf`. There are efforts being made in cargo to support per-crate build targets in workspaces (https://github.com/rust-lang/cargo/issues/7004), but the feature is unstable and doesn't quite work as expected yet.

I'm currently using this branch because I noticed that, without it, nix ends up building the library crate when I run the library tests, and building the library crate a second time when I build/run the application. When I use cargo directly, I use a shared `target` directory for all three crates so I only have to build the library crate once for each target.

While the limitation you mentioned is a little unfortunate, that is what I expected. It would be nice if the other application dependencies didn't have to get re-built if just the library crate changed, but for my current use-case I think that's acceptable.

Great point on the injection trick! I didn't think of using `overrideAttrs` to workaround this.

akiekintveld:
Just wanted to followup and see if you'd like to merge this PR, or if you'd prefer this use case be covered by the `overrideAttrs` workaround?

ipetkov:
Apologize for not following up on this earlier! I think given that `overideAttrs` makes for a good escape hatch I'd prefer we close this PR and keep `buildDepsOnly` a bit more opinionated at this time. Hopefully we'll improve the story for larger workspaces such that this dance won't be necessary in the future...

In the mean time, feel free to bring up other thoughts or improvements you may have in mind; thanks again for taking the time to open this PR!

PR #298, opened 2023-04-12
fix: do not install bindep artifacts by default

rvolosatovs:
## Motivation

1. Existing binary installation logic does not work with bindeps feature, since the binaries in the log are not necessarily the build outputs user expected to build, but rather just a dependency build. (I was previously resorting to parsing the list of binaries in the workspace and installing from that https://github.com/rvolosatovs/nixify/blob/3df7d15a90df4cf9ec57c6329b71f74eec4d47ac/lib/rust/mkAttrs.nix#L158-L176 https://github.com/rvolosatovs/nixify/blob/3df7d15a90df4cf9ec57c6329b71f74eec4d47ac/lib/rust/crateBins.nix#L21-L116)
2. ~It's not possible to build `dylib` artifacts, which may be useful in various contexts~ Done in #340
- ~Add support for installing crates utilizing `bindeps` feature by only installing binaries of workspace packages~
- ~Add support for installing `dylib` crate types~

## Checklist

- [x] updated `CHANGELOG.md`

### DIFF
--- a/lib/setupHooks/installFromCargoBuildLogHook.sh
+++ b/lib/setupHooks/installFromCargoBuildLogHook.sh
@@ -13,12 +13,17 @@ function installFromCargoBuildLog() (
   logs=$(jq -R 'fromjson?' <"${log}")
 
   local select_non_test='select(.reason == "compiler-artifact" and .profile.test == false)'
-  local select_bins="${select_non_test} | .executable | select(.!= null)"
+  local select_non_dep='select(contains("/deps/")| not)'
+  local select_bins="${select_non_test}"'| .executable | select(.!= null) | '"${select_non_dep}"
   local select_lib_files="${select_non_test}"'
-    | select(.target.kind | contains(["staticlib"]) or contains(["cdylib"]))
+    | select(.target.kind
+        | contains(["cdylib"])
+        or contains(["dylib"])
+        or contains(["staticlib"])
+    )
     | .filenames[]
     | select(endswith(".rlib") | not)
-  '
+    | '"${select_non_dep}"
 
   function installArtifacts() {
     local loc=${1?:missing}
@@ -32,12 +37,11 @@ function installFromCargoBuildLog() (
     rmdir --ignore-fail-on-non-empty "${loc}"
   }
 
-  echo "${logs}" | jq -r "${select_bins}" | installArtifacts "${dest}/bin"
-
   command cargo metadata --format-version 1 | jq '.workspace_members[]' | (
     while IFS= read -r ws_member; do
-      local select_member_libs="select(.package_id == ${ws_member}) | ${select_lib_files}"
-      echo "${logs}" | jq -r "${select_member_libs}" | installArtifacts "${dest}/lib"
+      local select_member="select(.package_id == ${ws_member})"
+      echo "${logs}" | jq -r "${select_member} | ${select_lib_files}" | installArtifacts "${dest}/lib"
+      echo "${logs}" | jq -r "${select_member} | ${select_bins}" | installArtifacts "${dest}/bin"
     done
   )
 
--- a/lib/setupHooks/installFromCargoBuildLogHook.sh
+++ b/lib/setupHooks/installFromCargoBuildLogHook.sh
@@ -12,10 +12,22 @@ function installFromCargoBuildLog() (
   local logs
   logs=$(jq -R 'fromjson?' <"${log}")
 
-  local select_non_test='select(.reason == "compiler-artifact" and .profile.test == false)'
-  local select_non_dep='select(contains("/deps/")| not)'
-  local select_bins="${select_non_test}"'| .executable | select(.!= null) | '"${select_non_dep}"
-  local select_lib_files="${select_non_test}"'
+  # We automatically ignore any bindeps artifacts as installation candidates.
+  # Note that if the same binary is built both as a bindep artifact for something else and as a
+  # "regular build", cargo will emit a log entry for each, meaning that we will never accidentally
+  # ignore installing a binary that the derivation was intending to build!
+  local select_non_deps_artifact='select(contains("/deps/artifact/") | not)'
+
+  # Only install binaries and libraries from the current workspace as a sanity check
+  local members="$(command cargo metadata --format-version 1 | jq -c '.workspace_members')"
+  local select_non_test_members='select(.reason == "compiler-artifact" and .profile.test == false)
+    | select(.package_id as $pid
+      | '"${members}"'
+      | contains([$pid])
+    )'
+
+  local select_bins="${select_non_test_members}| .executable | select(.!= null) | ${select_non_deps_artifact}"
+  local select_lib_files="${select_non_test_members}"'
     | select(.target.kind
         | contains(["cdylib"])
         or contains(["dylib"])
@@ -23,7 +35,7 @@ function installFromCargoBuildLog() (
     )
     | .filenames[]
     | select(endswith(".rlib") | not)
-    | '"${select_non_dep}"
+    | '"${select_non_deps_artifact}"
 
   function installArtifacts() {
     local loc=${1?:missing}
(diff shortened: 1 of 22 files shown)


lib/setupHooks/installFromCargoBuildLogHook.sh
```diff
@@ -13,12 +13,18 @@ function installFromCargoBuildLog() (
   logs=$(jq -R 'fromjson?' <"${log}")
 
   local select_non_test='select(.reason == "compiler-artifact" and .profile.test == false)'
-  local select_bins="${select_non_test} | .executable | select(.!= null)"
+  local select_non_dep='select((contains("/release/deps/") or contains("/debug/deps/"))| not)'
```
ipetkov, line 16: Should this check be shortened to just `select(contains("/deps/") | not)`? `release` and `dev` aren't the only profiles and additional custom profiles could trip this up

(Also, is there any other metadata that cargo emits in the json output to denote this as a build dep or is the path the only way to check for it?)

lib/setupHooks/installFromCargoBuildLogHook.sh
```diff
@@ -32,12 +38,11 @@ function installFromCargoBuildLog() (
...
 
-  echo "${logs}" | jq -r "${select_bins}" | installArtifacts "${dest}/bin"
-
   command cargo metadata --format-version 1 | jq '.workspace_members[]' | (
     while IFS= read -r ws_member; do
-      local select_member_libs="select(.package_id == ${ws_member}) | ${select_lib_files}"
-      echo "${logs}" | jq -r "${select_member_libs}" | installArtifacts "${dest}/lib"
+      local select_member="select(.package_id == ${ws_member})"
+      echo "${logs}" | jq -r "${select_member} | ${select_lib_files}" | installArtifacts "${dest}/lib"
+      echo "${logs}" | jq -r "${select_member} | ${select_bins}" | installArtifacts "${dest}/bin"
```
ipetkov, line 45: I think doing things this way will attempt to install _all binaries_ for a particular member, but that doesn't necessarily mean we have built them all. Consider if the derivation is built with `cargo build -p myCrate --bin foo` but the crate specifies `foo` and `bar` binaries. Would this check also attempt to install `bar`?

Perhaps we should be installing all binaries that were built _and are also not considered dependencies_?

lib/setupHooks/installFromCargoBuildLogHook.sh
```diff
@@ -13,12 +13,18 @@ function installFromCargoBuildLog() (
...
   local select_non_test='select(.reason == "compiler-artifact" and .profile.test == false)'
-  local select_bins="${select_non_test} | .executable | select(.!= null)"
+  local select_non_dep='select((contains("/release/deps/") or contains("/debug/deps/"))| not)'
+  local select_bins="${select_non_test}"'| .executable | select(.!= null) | '"${select_non_dep}"
   local select_lib_files="${select_non_test}"'
-    | select(.target.kind | contains(["staticlib"]) or contains(["cdylib"]))
+    | select(.target.kind
+        | contains(["cdylib"])
+        or contains(["dylib"])
+        or contains(["lib"])
```
ipetkov, line 22: Question: why do we want to install `lib` by default? According to [the reference](https://doc.rust-lang.org/reference/linkage.html) `lib` means 
> This is an ambiguous concept as to what exactly is produced because a library can manifest itself in several forms. The purpose of this generic lib option is to generate the "compiler recommended" style of library. The output library will always be usable by rustc, but the actual type of library may change from time-to-time.

I'm not entirely sure what this means or how it would be usable (but if it's something that can be used by other builds/linkers I don't mind including it; I'd just like to see an example to understand things better!)

lib/setupHooks/installFromCargoBuildLogHook.sh
```diff
@@ -13,12 +13,18 @@ function installFromCargoBuildLog() (
...
+        or contains(["rlib"])
```
ipetkov, line 23: Question: why do we want to install `rlibs` at all? These are intermediate artifacts used by the compiler and are generally (to my knowledge) not very useful outside of that. Plus the installed cargo artifacts dir should already include these (and in fact installing them all would largely result in installing the contents of cargo's `target` directory IIUC). I'd imagine if someone really needs those artifacts they should probably look for them there and not inside of `$out/lib`

lib/setupHooks/installFromCargoBuildLogHook.sh
rvolosatovs: I did not manage to find anything else than that, which is why I added the profile, just to make sure we're not grabbing something we should not (or we're in some weird edge case with e.g. nix store path set to `/deps` or something like that)

I'll remove the profile

lib/setupHooks/installFromCargoBuildLogHook.sh
rvolosatovs: I'd guess for some niche, very low-level use cases where the consumers of `crane` would want to e.g. build custom `cargoArtifacts`.
Happy to remove it

lib/setupHooks/installFromCargoBuildLogHook.sh
rvolosatovs: SGTM

lib/setupHooks/installFromCargoBuildLogHook.sh
ipetkov: Yeah it would definitely be easier to manage (and test!) with a Rust program. We already have `crane-utils` in the repo (currently only used for vendoring git-dependencies). Feel free to add a new binary in there which would process a cargo installation log!

lib/setupHooks/installFromCargoBuildLogHook.sh
ipetkov: Yeah I'd like to keep things as simple as possible for now, if someone has a use case which needs it we can revisit

Pegasust:
Would love to have this merged (I need `dylib`). Is there anything blocking this that I may take on to help?

lib/setupHooks/installFromCargoBuildLogHook.sh
rvolosatovs: Taking another look, this is actually already using `'"${select_non_dep}"` above, doesn't that cover the issue here?

rvolosatovs:
> Would love to have this merged (I need `dylib`). Is there anything blocking this that I may take on to help?

I'm just really swamped with work - rebased and addressed the feedback, don't have time to run the checks locally, so will (ab-)use CI for that, but I will pull in the newest version in my downstream project to test this.

I think all @ipetkov comments are addressed. @ipetkov if you agree, I'd propose filing an issue about a Rust program parsing the log. My intention was to port/adapt https://github.com/wasmCloud/wasmCloud/blob/934d70121e768eb33921bdaf5107881ba91ea963/tests/actors/build.rs#L13-L228, I was in fact, writing this code both thinking about how crane handles this and planning to contribute

ipetkov:
@rvolosatovs I'll try to take a closer look sometime this week, I've cached the details out of working memory and need to remember what my original comments meant or if they're still relevant :sweat_smile: (will play around with cargo's log format and see what data it gives us)

In the mean time I'm going to split out and land the `dylib` portion of the MR separately. I think that part is pretty uncontroversial and we can definitely land it as is, then focus on the bindeps portion more closely!

ipetkov:
@Pegasust feel free to track #340 for `dylib` support!

lib/setupHooks/installFromCargoBuildLogHook.sh
```diff
@@ -13,12 +13,17 @@ function installFromCargoBuildLog() (
   logs=$(jq -R 'fromjson?' <"${log}")
 
   local select_non_test='select(.reason == "compiler-artifact" and .profile.test == false)'
-  local select_bins="${select_non_test} | .executable | select(.!= null)"
+  local select_non_dep='select(contains("/deps/")| not)'
```
ipetkov, line 16: Played around with the bindeps feature myself and I found out that (fortunately) cargo will emit different log entries if a workspace target happens to _also_ be used as a bindep for another member. This means that we can always unconditionally ignore anything in `.../deps/artifact/...`, but we'll still install the actual targets if they happened to be part of the `cargo build` invocation, so modulo some comments here I'm happy to merge this in!

lib/setupHooks/installFromCargoBuildLogHook.sh
ipetkov: Looking back at my original comment, I think it might have come out of me misunderstanding the implementation so I'll mark this as resolved!

PR #299, opened 2023-04-15
Trunk support

### LINKED ISSUE #295: Trunk Support
[Trunk](https://trunkrs.dev/) is a tool to "Build, bundle & ship your Rust WASM application to the web"

I was able to use it with crane using the following:

```nix
let
  cargoTrunk =
    { cargoArtifacts
    , cargoTrunkExtraArgs ? ""
    , cargoTrunkExtraBuildArgs ? ""
    , cargoExtraArgs ? ""
    , ...
    }@origArgs:
    let
      args = builtins.removeAttrs origArgs [
        "cargoTrunkExtraArgs"
        "cargoTrunkExtraBuildArgs"
        "cargoExtraArgs"
      ];
    in
    craneLib.mkCargoDerivation (args // {
      inherit cargoArtifacts;
      pnameSuffix = "-trunk";

      buildPhaseCargoCommand = ''
        trunk ${cargoTrunkExtraArgs} build --release ${cargoTrunkExtraBuildArgs}
      '';

      installPhase = ''
        cp -r client/dist $out
      '';

      nativeBuildInputs = (args.nativeBuildInputs or [ ]) ++ (with pkgs; [
        trunk
        wasm-bindgen-cli
        binaryen
        nodePackages.sass # dart-sass compiled to javascript
      ]);
    });

  cargoArtifacts = craneLib.buildDepsOnly {
    inherit src buildInputs;
    cargoExtraArgs = "--target=wasm32-unknown-unknown --package=client";
    doCheck = false;
  };
in
  cargoTrunk {
    inherit buildInputs src cargoArtifacts;
    pname = "my-client";
    version = "0.1";
    cargoTrunkExtraBuildArgs = "client/index.html";
  }
```

I had to do some hacking to get it to work because we're using workspaces.
Perhaps this could be cleaned up and turned into a officially supported function?

lelgenio:
## Motivation

[Trunk](https://trunkrs.dev) is an incredibly usefull tool to "Build, bundle & ship your Rust WASM application to the web".

Closes #295 

## Checklist

- [x] updated `CHANGELOG.md`

## TODO:
- [x] Add an example of how to build a workspace with three crates: "client", "server" and "shared"
- [ ] Perhaps bring the documentation more in line with that of other functions (?)
- [x] Squash


docs/API.md
```diff
@@ -226,6 +226,25 @@ The following hooks are automatically added as native build inputs:
...
 * `removeReferencesToVendoredSourcesHook`
 
+### `craneLib.buildTrunkPackage`
+`buildTrunkPackage :: set -> drv`
+
+Create a derivation which will build a distributable directory for a WASM application.
+
+Except where noted below, all derivation attributes are delegated to
+`mkCargoDerivation`, and can be used to influence its behavior.
+* `cargoArtifactsWasm`: A path (or derivation) which contains an existing cargo
```
ipetkov, line 236:
```suggestion
* `cargoArtifacts`: A path (or derivation) which contains an existing cargo
```

docs/API.md
```diff
@@ -226,6 +226,25 @@ The following hooks are automatically added as native build inputs:
...
+  `target` directory, which will be reused at the start of the derivation.
+  Useful for caching incremental cargo builds.
+  - This can be prepared via `buildDepsOnly`
+  - Alternatively, any cargo-based derivation which was built with
+    `doInstallCargoArtifacts = true` will work as well
+  - For this to be usefull it must have the dependencies
+    built for the `wasm32-unknown-unknown` target
+* `indexPath` A path to the index.html of your trunk project
```
ipetkov, line 244:
```suggestion
* `trunkIndexPath` A path to the index.html of your trunk project
```

lib/buildTrunkPackage.nix
```diff
@@ -0,0 +1,44 @@
...
+, wasm-bindgen-cli
+, binaryen
+, nodePackages
+}:
+
+{
+  # this is not called cargoArtifacts to push users
+  # away from giving us useless artifacts built
+  # for native target instead of wasm32-unknown-unknown
+  cargoArtifactsWasm
```
ipetkov, line 12:
```suggestion
  cargoArtifacts
```

I sympathize with the intent but I'd prefer we call this `cargoArtifacts` since that is the pattern we already use everywhere else (plus various low level helpers know and look for that name).

Yes it is a slight foot gun if someone passes in `cargoArtifacts` related to another target, though it's totally possible to have `cargoArtifacts` consisting of multiple targets too. I'd rather be consistent here and revisit if problems arise

lib/buildTrunkPackage.nix
```diff
@@ -0,0 +1,44 @@
...
+, indexPath ? "./index.html"
+, trunkExtraArgs ? ""
+, trunkExtraBuildArgs ? ""
```
ipetkov, line 15:
```suggestion
, trunkExtraArgs ? ""
, trunkExtraBuildArgs ? ""
, trunkIndexPath ? "./index.html"
```

Should we call this `trunkIndexPath` for consistency with the other `trunk*` parameters?

ipetkov (commented):
Thank you for the comprehensive PR, I really appreciate that you not only added tests but also added API docs and examples!

lib/buildTrunkPackage.nix
lelgenio: Having a second pass on the PR now, how about having an assert in `buildTrunkPackage` checking if `${cargoArtifacts}/target/wasm32-unknown-unknown` is a path?

lib/buildTrunkPackage.nix
lelgenio: Makes sense to me!

lelgenio:
I'm not sure about it, but it seems like a good idea to generate `cargoArtifacts` if they are not passed in, considering `buildTrunkPackage` is supposed to behave similarly to `buildPackage`, is that right?

lib/buildTrunkPackage.nix
```diff
@@ -0,0 +1,63 @@
...
+    "trunkExtraArgs"
+    "trunkExtraBuildArgs"
+  ];
+
+  incorrectTargetError = throw "cargoArtifacts must be built for target wasm32-unknown-unknown";
+  builtTargets = builtins.readDir "${cargoArtifacts}/target";
+  wasmTargetDir = builtTargets."wasm32-unknown-unknown" or incorrectTargetError;
+in
+
+assert cargoArtifacts == null || wasmTargetDir == "directory";
```
ipetkov, line 31: Rather than doing this assertion I think a better alternative is to set `CARGO_BUILD_TARGET = "wasm32-unknown-unknown"` as a default (if not overwritten)? That would cover most of the common cases (esp with adding a default call to `cargoArtifacts = buildDepsOnly ...` when not set too) while avoiding IFD headaches or assumptions about the artifact format (since it can be customized)

ipetkov:
> I'm not sure about it, but it seems like a good idea to generate `cargoArtifacts` if they are not passed in, considering `buildTrunkPackage` is supposed to behave similarly to `buildPackage`, is that right?

I like this idea!

lib/buildTrunkPackage.nix
lelgenio: Yes this is much better, using this less intrusive env var never crossed my mind.

lelgenio:
CI fail on 22.11 is due to `wasm-bindgen-cli` beign outdated. Made a backport PR https://github.com/NixOS/nixpkgs/pull/228754.

ipetkov:
Pushed a few minor fixes, will merge when CI passes!

PR #314, opened 2023-05-02
fix: building trunk projects that use sass, add test

lelgenio:
## Motivation

The current format for `TRUNK_TOOLS_SASS` is incorrect from Trunk's perspective, it expects the entire first line from the command `sass --version`, this is no problem with the native version of `dart-sass` which outputs only the version `1.56.0`. But the version currently in use is compiled to javascript and has aditional information `1.56.0 compiled with dart2js 2.18.4`.

This PR fixes the proccessing done to `TRUNK_TOOLS_SASS` to match what is expected from Trunk. It also tests sass compilation, and adds a small sass file to the trunk examples.

Thanks to this comment for highligthing the issue: https://github.com/ipetkov/crane/issues/295#issuecomment-1532096291.

## Checklist

- [ ] updated `CHANGELOG.md`

### DIFF
--- a/checks/trunk.nix
+++ b/checks/trunk.nix
@@ -30,6 +30,7 @@ let
 in
 runCommand "trunkTests" { } ''
   test -f ${trunkSimple}/*.wasm
+  test -f ${trunkSimple}/*.css
   test -f ${trunkSimpleNoArtifacts}/*.wasm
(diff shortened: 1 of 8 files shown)


ipetkov:
Thanks for fixing this!

Issue #76, opened 2022-08-14
A way to output just a diff of `./target`


dpc:
I was admiring the results of `crane`-based CI pipeline in Fedimint: https://github.com/fedimint/fedimint/actions/runs/2853966200 . After adding all the proper `src` filtering, when no source files were modified, nothing gets rebuild - not even tests are being re-run (since they've already passed successful, there's no point in re-running them).

Which is frankly super awesome, and thank you so much for working on `crane`! I can't overstate how well it works and how helpful you are. :)

However as you can see the build still takes ~2m, and it's all downloading stuff from cachix. There is something off there - as 2GB of stuff shouldn't take 2 minutes to download, IMO, but that's something I'm going to investigate on my own.

I started looking into why is it 2GB of data, and I realized that part of the reason is that build contains two `target/` outputs:

```
Cachix: push
  /home/ubuntu/actions-runner/_work/_actions/cachix/cachix-action/v10/dist/main/push-paths.sh /home/ubuntu/.nix-profile/bin/cachix fedimint 
  compressing and pushing /nix/store/08mljabqrb8p11dl3y9ddqbakiss3mhl-source (108.28 KiB)
  compressing and pushing /nix/store/gw7pv5j6djw3rmkxvv9lj5az7l7xccfk-cargo-package-0.0.1 (489.87 MiB)
  compressing and pushing /nix/store/h1riypzjkqxc6qvw0m6f0kw7dvnrhai5-source (744.02 KiB)
  compressing and pushing /nix/store/acpnlx92cd7rgsm028l1g4mmf1575c4g-source (760.59 KiB)
  compressing and pushing /nix/store/9d3fwbq3bp0n03lrbs6msmacj2m9y7ml-cargo-package-0.0.1 (96.00 B)
  compressing and pushing /nix/store/glzdgkib6v3mf556vfhb7zy6ziqadk1l-cargo-package-0.0.1 (96.00 B)
  compressing and pushing /nix/store/9gmk58528m11yrra4miapcca0wv6ywnc-cargo-package-0.0.1 (420.72 MiB)
  compressing and pushing /nix/store/qdwmpys2zsr3klx1p92c730zw7jn0ali-cargo-package-0.0.1 (96.00 B)
  compressing and pushing /nix/store/qph9acr5h18gj51g1hv9d56ahngnxqym-cargo-package-0.0.1 (96.00 B)
[...]

PR #322, opened 2023-05-11
Symlink cargo build artifacts instead of copying them

Philipp-M:
## Motivation

Avoid copying big cargo build artifacts from the nix store, when using `cargoArtifacts`.
This works by just symlinking everything in `target/release/build` and `target/release/deps`. and uses rsync to copy other files. AFAIK the big files are all contained in these folders and are content-addressed by hash, so this *should* be safe.

What I have noticed (even before this change), is that cargo tarpaulin generally seems to rebuild the cargo artifacts, ignoring previous artifacts, even with `--skip-clean`, this fails as it tries to regenerate already built artifacts and tries to write into the build folders.
This should probably be investigated, so that cargo tarpaulin reuses the previously built/cached dependencies.
I have for now just removed `--skip-clean` in cargo tarpaulin, as it doesn't seem to have the intended effect anyway?
Without that flag, all tests run through.

Btw. I have noticed the (now removed) comment, but I have not experienced any issues other than the cargo tarpaulin above.

## Checklist

- [ ] updated `CHANGELOG.md`

### DIFF
--- a/lib/setupHooks/inheritCargoArtifactsHook.sh
+++ b/lib/setupHooks/inheritCargoArtifactsHook.sh
@@ -24,18 +24,36 @@ inheritCargoArtifacts() {
   elif [ -d "${preparedArtifacts}" ]; then
     echo "copying cargo artifacts from ${preparedArtifacts} to ${cargoTargetDir}"
 
-    # NB: rustc doesn't like it when artifacts are either symlinks or hardlinks to the store
-    # (it tries to truncate files instead of unlinking and recreating them)
-    # so we're forced to do a full copy here :(
-    #
-    # Notes:
-    # - --no-target-directory to avoid nesting (i.e. `./target/target`)
-    # - preserve timestamps to avoid rebuilding
-    # - no-preserve ownership (root) so we can make the files writable
-    cp -r "${preparedArtifacts}" \
-      --no-target-directory "${cargoTargetDir}" \
-      --preserve=timestamps \
-      --no-preserve=ownership
+    # copy target dir but ignore crate build artifacts
+    rsync -r --chmod=Du=rwx,Dg=rx,Do=rx --exclude "release/build" --exclude "release/deps" --exclude "*/release/build" --exclude "*/release/deps" "${preparedArtifacts}/" "${cargoTargetDir}/"
+
+    link_build_artifacts() {
+      local artifacts="$1"
+      local target="$2"
+
+      if [ -d "${artifacts}/release/deps" ]; then
+        mkdir -p "${target}/release/deps"
+        for dep in $(ls "${artifacts}/release/deps"); do
+          ln -fs "${artifacts}/release/deps/$dep" "${target}/release/deps/$dep"
+        done
+      fi
+
+      if [ -d "${artifacts}/release/build" ]; then
+        mkdir -p "${target}/release/build"
+        for build in $(ls "${artifacts}/release/build"); do
+          ln -fs "${artifacts}/release/build/$build" "${target}/release/build/$build"
+        done
+      fi
+    }
+
+    # symlink crate build artifacts
+    link_build_artifacts "${preparedArtifacts}" "${cargoTargetDir}"
+
+    # for each build target as well
+    # all other directories are ignored in `link_build_artifacts`
+    for target in $(ls "${preparedArtifacts}"); do
+      link_build_artifacts "${preparedArtifacts}/$target" "${cargoTargetDir}/$target"
+    done
 
     # Keep existing permissions (e.g. exectuable), but also make things writable
     # since the store is read-only and cargo would otherwise choke
--- a/lib/cargoTarpaulin.nix
+++ b/lib/cargoTarpaulin.nix
@@ -3,7 +3,7 @@
 }:
 
 { cargoExtraArgs ? ""
-, cargoTarpaulinExtraArgs ? "--skip-clean --out Xml --output-dir $out"
+, cargoTarpaulinExtraArgs ? "--out Xml --output-dir $out"
 , ...
 }@origArgs:
 let
--- a/lib/setupHooks/inheritCargoArtifacts.nix
+++ b/lib/setupHooks/inheritCargoArtifacts.nix
@@ -1,8 +1,10 @@
 { makeSetupHook
+, rsync
 }:
 
 makeSetupHook
 {
   name = "inheritCargoArtifactsHook";
+  propagatedBuildInputs = [ rsync ];
 } ./inheritCargoArtifactsHook.sh
 


dpc:
Have you seen #76 ?

dpc:
I gave it a spin, but realized it can't help me as I'm running with:

```
installCargoArtifactsMode = "use-zstd";
```

for reasons explained in #76.

I turned it off, but then discovered that after building dependencies Nix does bunch of `shrinking ...` on lots of files, and then it causes building the workspace to rebuild all the dependencies again. :/

I'm not sure if `master` has some error, or my setup is triggering some issues (works OK with `use-zstd`).

So I can't tell how much does it improve things.

dpc:
It seems like the problem is not there in `v0.12.1`, so I rebased your commit on top of `v0.12.1`:

```
9ca63fb40f4ff761a6cbb7d5c07f5fe4c98782b9 (HEAD -> symlink-artifacts, origin/symlink-artifacts) Symlink cargo build artifacts instead of copying them
445a3d222947632b5593112bb817850e8a9cf737 (tag: v0.12.1) Update CHANGELOG
```

and again I'm seeing rebuilds of dependencies. So either your change is causing it too, or I messed up something somewhere.

dpc:
I tried again (to double check) just `v0.12.1` and it doesn't have the problem.

dpc:
I'm using fenix toolchain. I reported https://github.com/ipetkov/crane/issues/323 , maybe it will shed some light.

ipetkov:
Hi @Philipp-M thank you for the PR! When I originally explored symlinking all artifacts (instead of copying them deeply) cargo would fail with some errors about trying to write to a read-only file system (i.e. the nix store), but I hadn't considered that this might not apply to _all_ artifacts, so kudos to spotting that!

---

> What I have noticed (even before this change), is that cargo tarpaulin generally seems to rebuild the cargo artifacts, ignoring previous artifacts

I don't use `cargo-tarpaulin` much these days, but I do remember it insisting on cleaning the target directory (to recompile all deps with the right compiler flags IIRC). If it is ignoring `--skip-clean` it deserves some investigation, but it could be an unrelated issue

---

Noting @dpc 's comment on #323 (please correct me if I'm misunderstanding or misrepresenting anything)

> Edit: Just checked and it seems that v0.12.1 behaves correctly. It also does the shrinking but when buildWorkspace starts it only builds local crates as expected.

Sounds like these changes (as they stand currently) are leading to rebuilds. My first guess is that cargo may be getting confused by looking at artifacts whose timestamp is older than the sources (sources in the Nix store have a timestamp in 1970) and it is resulting in rebuilding them. I wonder if the artifacts should be _hardlinked_ (not symlinked) so that they get their own timestamp instead of inheriting the previous one...

lib/setupHooks/inheritCargoArtifactsHook.sh
```diff
@@ -24,18 +24,36 @@ inheritCargoArtifacts() {
...
+        done
+      fi
+
+      if [ -d "${artifacts}/release/build" ]; then
+        mkdir -p "${target}/release/build"
+        for build in $(ls "${artifacts}/release/build"); do
+          ln -fs "${artifacts}/release/build/$build" "${target}/release/build/$build"
+        done
+      fi
+    }
```
ipetkov, line 47: One thing worth noting is that this bit of logic is assuming that only `release` artifacts are being built. Although this is the default build behavior we apply, it's entirely possible a derivation can be configured to build any other profile (including custom ones).

Also the derivation could also be building for different targets (though I believe this is captured by the wildcards). I think a more robust approach might be to find out exactly which files cargo will tolerate as symlinks/hardlinks and use that instead (e.g. `*.rlib`, `*.rmeta` or `*/build/*`, etc.)

dpc: Ohhh. I 99% was testing it with `debug` build, as our crane-based derivations default to debug-inherited profile. That would explain the rebuilds.

Philipp-M: > I think a more robust approach might be to find out exactly which files cargo will tolerate as symlinks/hardlinks and use that instead (e.g. *.rlib, *.rmeta or */build/*, etc.)

After a little bit of research, I think it's safer to just symlink the content-addressed artifacts (all files under `build` and `deps`), these are the artifacts that take most (almost all the) disk-space and I had some weird issues when testing something like this (create dirs recursively and then link *every* file from the nix store cargo target to the build target dir):

```bash
    find "${preparedArtifacts}" -type d -printf '%P\n' | sed '/^$/d' | xargs -d '\n' -n 100 -P 10 mkdir -p
    find "${preparedArtifacts}" -type f -printf '%P\n' | sed '/^$/d' | xargs -P 100 -d '\n' -I '##{}##' ln -fs "${preparedArtifacts}/##{}##" "##{}##"
```
Some `.d` files in the root of the build-target could not be overwritten, but others were overwritten...

I have found a way to make the code a little bit more concise and performant (symlinking of the amount of files can take some time, not much, but if it can be faster it should be faster :)) and now every build target and profile should be considered.

@dpc, can you try the latest commit?

ipetkov: In the spirit of going as faster to be faster, we could even invoke `mkdir -p` on all directories with no child directories so we can do it in fewer spawned processes?

```
find "${preparedArtifacts}" -type d -links 2 ! -empty -printf '%P\n' | sed '/^$/d' | xargs -d '\n' -n 100 -P 10 mkdir -p
```

dpc: @Philipp-M Everything failing on c41599656db25662504c561f5c2b381e29621095

ipetkov:
@Philipp-M I pushed a small fix to resolve the build errors. Looks like a lot of the tests are passing but some are still failing (perhaps we shouldn't link build script output directories since scripts might need to be re-run and they expect to be able to write files? Alternatively perhaps we can take a shortcut by linking any `*.rlib` files or similar heuristics rather than guessing which directories are for build scripts and which ones aren't)

Philipp-M:
Thanks.
Sorry haven't had time to investigate this further (since it's also a little bit hard to debug, as on my machines everything seems to be working). According to the tests only OSX has problems (though I guess this has to be tested thoroughly on different machines anyway I think). I guess to be sure, one has to deep dive into the source of how cargo etc. is deciding whether stuff has to be rebuilt on different machines. Leaving build script dirs out of the linking process, is probably not a bad idea:

> When rebuilding a package, Cargo does not necessarily know if the build script needs to be run again. By default, it takes a conservative approach of always re-running the build script if any file within the package is changed
> ...
> The rerun-if-changed instruction tells Cargo to re-run the build script if the file at the given path has changed. Currently, Cargo only uses the filesystem last-modified “mtime” timestamp to determine if the file has changed. It compares against an internal cached timestamp of when the build script last ran.
> 
> If the path points to a directory, it will scan the entire directory for any modifications.

@dpc, can you provide more info, what exactly is going wrong (like hardware/OS, the exact issue)?

Philipp-M:
Just noticed #344 (thanks for further investigating btw.), I guess I can close this for now?
One thing I noticed while checking the `deps` folder, is that I think `.so` files can be safely symlinked as well.

ipetkov:
@Philipp-M It's up to you! I decided to pick the low hanging fruit so we could make incremental updates. 

Feel free to explore the .so linking optimization in this PR or a fresh one as you prefer

dpc:
I tried the current master with #334 and just so you're aware: in a larger project, just using zstd is still waaay better than not using, even with current optimizations.

In our our project zstd produces: 828M for deps, and 1.5G for workspace build. no-zstd&symlinks: 4.6G and 4.1G.

ipetkov:
@dpc fwiw this PR's symlinking only happens when _inheriting existing artifacts_ to perform a build, but is entirely unrelated with how the _resulting artifacts_ are organized in `/nix/store`

PR #324, opened 2023-05-14
Natively support sparse registries

j-baker:
## Motivation

With Cargo 1.68.0, sparse registries were made stable. With sparse registries, index metadata is stored in an HTTP api rather than in a git repository.

As relevant to Crane, the relevant changes are mostly that registries do not always start with `registry+` and mostly start with `registry+` or `sparse+` depending on whether the registry is sparse or not.

This PR adjusts the core of Crane to differentiate between `registry` and `sparse`, and adds a new sparse registry factory to make things easy.

I'm new to Nix, so it's possible and likely that my code is incorrect.

## Checklist

- [X] added tests to verify new behavior
- [X] added an example template or updated an existing one
- [X] updated `docs/API.md` (or general documentation) with changes
- [X] updated `CHANGELOG.md`

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -36,11 +36,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1683020768,
-        "narHash": "sha256-ZyZl6k9NWS5QPwD3NoAVz/eSgodQDvl+y+fu8MVbrHc=",
+        "lastModified": 1684120848,
+        "narHash": "sha256-gIwJ5ac1FwZEkCRwjY+gLwgD4G1Bw3Xtr2jr2XihMPo=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "44f30edf5661d86fb3a95841c35127f3d0ea8b0f",
+        "rev": "0cb867999eec4085e1c9ca61c09b72261fa63bb4",
         "type": "github"
       },
       "original": {
@@ -68,11 +68,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1683080331,
-        "narHash": "sha256-nGDvJ1DAxZIwdn6ww8IFwzoHb2rqBP4wv/65Wt5vflk=",
+        "lastModified": 1684117262,
+        "narHash": "sha256-ZSF4CZqeyk6QwTjal73KPMuTWiU6w/p8ygEimrPb7u4=",
         "owner": "oxalica",
         "repo": "rust-overlay",
-        "rev": "d59c3fa0cba8336e115b376c2d9e91053aa59e56",
+        "rev": "4679872d2dd3e94ffef75efcbf77ea11549d90a7",
         "type": "github"
       },
       "original": {
--- a/lib/registryFromSparse.nix
+++ b/lib/registryFromSparse.nix
@@ -1,14 +1,15 @@
-{ registryFromDownloadUrl }:
+{ registryFromDownloadUrl, lib }:
 
 { indexUrl
-, sha256
+, configSha256
 , fetchurlExtraArgs ? { }
 }:
 
 let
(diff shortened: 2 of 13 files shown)


lib/registryFromSparse.nix
```diff
@@ -0,0 +1,23 @@
+{ registryFromDownloadUrl }:
+
+{ indexUrl
+, sha256
```
ipetkov, line 4:
```suggestion
, configSha256
```

Nit: I'd prefer to name this `configSha256` to highlight that this is the hash of the _configuration file_ and not just the contents of the root URL or something!

lib/registryFromSparse.nix
```diff
@@ -0,0 +1,23 @@
...
+, fetchurlExtraArgs ? { }
+}:
+
+let
+  configContents = builtins.readFile "${builtins.fetchurl {
+    inherit sha256;
+    url = "${indexUrl}config.json";
```
ipetkov, line 11:
```suggestion
    url = "${indexUrl}/config.json";
```

Nit: I'd prefer we include the `/` here rather than assume the user will specify a url with a trailing slash

lib/urlForCargoPackage.nix
```diff
@@ -19,8 +19,10 @@ let
     else if nameLen == 3 then "3/${substr 0 1 name}"
     else "${substr 0 2 name}/${substr 2 4 name}";
 
+  knownRegistries = lib.concatStringsSep " " (builtins.attrNames crateRegistries);
   registry = crateRegistries.${source} or (throw ''
     not sure how to download crates from ${source}.
+    known registries are ${knownRegistries}.
```
ipetkov, line 25:
```suggestion
  knownRegistries = lib.concatStringsSep "\n" (builtins.attrNames crateRegistries);
  registry = crateRegistries.${source} or (throw ''
    not sure how to download crates from ${source}.
    known registries are:
    ${knownRegistries}
    
```

Might be better to print these out on one line each for easier scanning

ipetkov (commented):
Thanks for the PR @j-baker, this is looking great! Just have some various nits/clean up comments

lib/registryFromSparse.nix
j-baker: ack!

flake.lock
```diff
@@ -36,11 +36,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1683020768,
```
j-baker, line 39: I updated the flake lockfile to ensure that Cargo version is >= 1.68.

lib/registryFromSparse.nix
j-baker: I did that - here, I remove the trailing slash for getting the config json, and additionally i handle it in the 'fetch from download url' case.

flake.lock
ipetkov: We also test against the latest NixOS release (new release comes out at the end of the month). Also worth noting that sparse registires are getting fully stabilized in 1.70 which also comes out in two weeks

ipetkov:
@j-baker I had originally assumed the build issues were due to using an older rustc, but after merging the latest `master` in, I can't seem to get the tests to pass locally

```
error: not sure how to download crates from registry+https://github.com/rust-lang/crates.io-index.
known registries are:
registry+https://github.com/rust-lang/crates.io-index/
```

I have a feeling that something about the trailing slash handling on URLs is tripping the evaluation up (`registry+` URLs do not have a trailing slash in `Cargo.lock` so maybe that's related)

j-baker:
Hi, ivan! I discovered the same thing this week and had to use an older
commit. I think when fixing PR comments i broke the crates.io case. I’ll
take a look early next week.

ipetkov:
I think I figured out the issue! Let's see if the CI passes

PR #327, opened 2023-05-18
Keep "resolver" when cleaning Cargo.toml

Detegr:
For projects using version != 2021 and resolver = 2, the resolver information needs to be kept or otherwise the dependency build might fail. It doesn't hurt to always keep the information, so it's now always included in the cleaned Cargo.toml.

I created a minimal repro of the issue [here](https://github.com/Detegr/crane-resolver-repro).

## Checklist

- [x] updated `CHANGELOG.md`


ipetkov (approved):
LGTM, thanks!

PR #330, opened 2023-05-23
fix: Strip store references in buildTrunkPackage

lelgenio:
## Motivation

The output of `buildTrunkPackage` should never have runtime dependencies. 
Not only are they redundant but they also increase the closure greatly.
This makes a big difference when using nix to build docker images (from 64Mb to 5Mb).

## Checklist

- [x] updated `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 * Added support for the [Trunk](https://trunkrs.dev) wasm app build tool
 
+### Fixed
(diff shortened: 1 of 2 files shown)


ipetkov (approved):
Thanks for fixing this!

Pushed a minor typo fix, will merge when CI is green!

lelgenio:
Why does CI for macos take so long? 🤔

ipetkov:
There's a cross compilation test if and if the artifacts aren't cached it takes forever to build. I haven't figured out a good way to pin the examples just for CI such that this doesn't happen as much ¯\_(ツ)_/¯

PR #334, opened 2023-06-02
Clean install hooks from buildDepsOnly invocation

uosis:
## Motivation

Currently, install hooks get passed down to `buildDepsOnly` call, and, if they need any source files not included in the dummy source, build fails.

It seems reasonable to clean them together with custom install commands.

## Checklist

- [ ] updated `CHANGELOG.md`

### DIFF
--- a/lib/buildPackage.nix
+++ b/lib/buildPackage.nix
@@ -42,7 +42,7 @@ mkCargoDerivation (cleanedArgs // memoizedArgs // {
         installCargoArtifactsMode = args.installCargoArtifactsMode or "use-zstd";
       };
     in
-    buildDepsOnly (removeAttrs depsArgs [ "installPhase" "installPhaseCommand" ])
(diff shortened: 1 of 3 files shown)


lib/buildPackage.nix
```diff
@@ -42,7 +42,7 @@ mkCargoDerivation (cleanedArgs // memoizedArgs // {
         installCargoArtifactsMode = args.installCargoArtifactsMode or "use-zstd";
       };
     in
-    buildDepsOnly (removeAttrs depsArgs [ "installPhase" "installPhaseCommand" ])
+    buildDepsOnly (removeAttrs depsArgs [ "installPhase" "installPhaseCommand" "preInstall" "postInstall" ])
```
ipetkov, line 45: Perhaps instead of trying to clean these up we should just set `installPhase = "mkdir -p $out"` in `depsArgs` which will _not_ run any installation hooks and solve the issue a bit more simply!

PR #338, opened 2023-06-06
fix(Trunk): Set the `wasm-opt` version as expected by Trunk

tinou98:
Trunk expect a version number with this specific format: "version_{num}"
(see https://github.com/thedodd/trunk/blob/master/src/tools.rs#L92)

## Motivation

When using `<link data-trunk rel="rust" data-wasm-opt="4" />` `buildTrunkPackage` fails while trying to download `wasm-opt`

## Checklist

- [X] updated `CHANGELOG.md`

For the test part, we can add the `data-wasm-opt="4"` to one of the existing test, but I'm waiting your input before doing so.


ipetkov:
This looks good to me!

cc @lelgenio as FYI

Issue #354, opened 2023-07-06
How to influence wasm-bindgen-cli version


HBBrennan:
Hello, I have a rust project that is nearly identical to https://crane.dev/examples/trunk-workspace.html. When doing a nix build, I see get an error with a wasm-bindgen ABI version mismatch. Specifically:

```
configuring trunk tools
TRUNK_TOOLS_SASS=1.63.3 compiled with dart2js 3.0.4
TRUNK_TOOLS_WASM_BINDGEN=0.2.84
TRUNK_TOOLS_WASM_OPT=version_112
```
...
```
2023-07-06T22:43:26.680486Z  INFO using system installed binary app=wasm-bindgen version=0.2.84
2023-07-06T22:43:26.680589Z  INFO calling wasm-bindgen for client
error:

it looks like the Rust project used to create this wasm file was linked against
version of wasm-bindgen that uses a different bindgen format than this binary:

  rust wasm file schema version: 0.2.87
     this binary schema version: 0.2.84

Currently the bindgen format is unstable enough that these two schema versions
must exactly match. You can accomplish this by either updating the wasm-bindgen
dependency or this binary.
```

ipetkov:
Hi @HBBrennan!

There's two ways that come to mind:
1. Update your `nixpkgs` input (or whatever instantiation of `pkgs` is passed into `crane.mkLib`) to use `wasm-bindgen` 0.2.87 (looks like the latest version of `wasm-bindgen-cli` is currently 0.2.84 in `nixpkgs-unstable`, you might need to apply your own patch/overlay or contribute the change upstream)
2. Downgrade your entire dependency tree to using `wasm-bindgen` 0.2.84 via `cargo update -p wasm-bindgen --precise 0.2.84` (and if that doesn't work, `cargo tree -i wasm-bindgen@0.2.87` may help you figure out what is pulling it in)

Hope this helps!

HBBrennan:
Thank you so much I'll unblock myself with the latter. Love the project btw.

jhartma:
Had the same problem and fixed it with ```wasm-bindgen = "= 0.2.84"```in Cargo.toml

rambip:
Same problem here.
This really needs a fix !
If someone left `wasm-bindgen = "=0.2.84"` in her Cargo.toml, it will fail to build when the wasm-bindgen version change in nixpkgs.

[...]

PR #363, opened 2023-08-08
buildTrunkPackage: Add trunkWasmBindgenCliPkg argument

lelgenio:
I named the argument `trunkWasmBindgenCliPkg`, hinting to a future `trunkWasmBindgenCliVersion`, but that's another story.

The warning could be improved I think. Or maybe it should be an error?

This is blocked by: https://github.com/NixOS/nixpkgs/pull/247270

## Motivation

Should help with https://github.com/ipetkov/crane/issues/354

## Checklist

- [x] updated `CHANGELOG.md`
- [ ] Figure out a good name for the argument.
- [ ] Improve warning message.
- [ ] Figure out if it should be a warning or an error.


ipetkov:
I think at this rate we might as well name the parameter `wasm-bindgen-cli` (but have it default to the version in `pkgs` if not specified) which gives us the best of both worlds (and makes it easy to `inherit (pkgs) wasm-bindgen-cli` without having to use a separate variable name)

rambip:
Is it planned that `crane` will store the hashes of each version of wasm-bindgen-cli ?

That would be really convenient, getting the hash each time I have to change my version of wasm-bindgen-cli is a pain ...

rambip:
You can see how I did it [here](https://github.com/rambip/wasm-tooling/blob/main/update-wasm-bindgen.sh)

ipetkov:
> Is it planned that `crane` will store the hashes of each version of wasm-bindgen-cli ?

I don't know if `crane` would be the most appropriate place for this or if it should go into some kind of related project/repo

rambip:
> > Is it planned that `crane` will store the hashes of each version of wasm-bindgen-cli ?
> 
> I don't know if `crane` would be the most appropriate place for this or if it should go into some kind of related project/repo

Since there doesn't seem to be most appropriate projects, I do think it would be nice to have it in crane.

checks/trunk-outdated-bindgen/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "trunk"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
```
ipetkov, line 6: Should this pin `wasm-bindgen` to 0.2.85 so that we can test that things keep working?

Otherwise the next time we run `cargo update` in here it might break the test

ipetkov (approved):

checks/trunk-outdated-bindgen/Cargo.toml
lelgenio: Done :+1:

ipetkov:
Thanks for addressing this!

Issue #366, opened 2023-08-11
end-to-end test example


lelgenio:
One thing I've found to be pretty useful is running E2E tests as a nix derivation. I'm thinking of cleaning up and contributing this as an example, but requesting some comments before.

Maybe this sort of thing is best left to a separate project?

```nix
{
  description = "Example E2E testing";
  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/release-23.05";

    crane = {
      url = "github:ipetkov/crane";
      inputs = {
        nixpkgs.follows = "nixpkgs";
        rust-overlay.follows = "rust-overlay";
        flake-utils.follows = "flake-utils";
      };
    };

    rust-overlay = {
      url = "github:oxalica/rust-overlay";
      inputs = {
        nixpkgs.follows = "nixpkgs";
        flake-utils.follows = "flake-utils";
      };
    };

    flake-utils.url = "github:numtide/flake-utils";
  };
  outputs = { self, nixpkgs, crane, rust-overlay, flake-utils, ... }:
    flake-utils.lib.eachDefaultSystem (system:
      let
        pkgs = import nixpkgs {
          inherit system;
        };
        inherit (pkgs) lib;
        craneLib = crane.lib.${system};
        src = craneLib.cleanCargoSource (craneLib.path ./.);

        buildInputs = with pkgs; [ pkg-config openssl ];

        commonArgs = {
          inherit src buildInputs;
          pname = "example-e2e";
          version = "0.1";
          doCheck = false;
        };

        cargoArtifacts = craneLib.buildDepsOnly commonArgs;

        server = craneLib.buildPackage (commonArgs // {
          inherit cargoArtifacts;
          cargoExtraArgs = "--package=server";
        });
        e2eTester = craneLib.buildPackage (commonArgs // {
          inherit cargoArtifacts;
          cargoExtraArgs = "--package=e2e_tests";
        });

        runE2ETests = pkgs.runCommand "e2e-tests" { } ''
          HOME="$(mktemp -d)"

          wait-online() {
            timeout 5s \
              ${pkgs.retry}/bin/retry --until=success --delay "1,1,1,1,1" -- \
                ${pkgs.curl}/bin/curl -s "$@"
          }

          ${pkgs.postgresql}/bin/initdb postgres-data
          ${pkgs.postgresql}/bin/pg_ctl --pgdata=postgres-data --options "-c unix_socket_directories=$PWD" start
[...]

PR #368, opened 2023-08-12
Add example of E2E testing

lelgenio:
## Motivation

Discussion: https://github.com/ipetkov/crane/issues/366

## Checklist

- [ ] updated `examples/README.md`


examples/end-to-end-testing/flake.nix
```diff
@@ -0,0 +1,85 @@
...
+          ${workspace}/bin/server &
+          wait-for-connection --fail localhost:8000
+
+          # Firefox likes to write to $HOME
+          HOME="$(mktemp -d)"
+          geckodriver &
+          wait-for-connection localhost:4444
+
+          export SSL_CERT_FILE="${pkgs.cacert}/etc/ssl/certs/ca-bundle.crt"
+          ${workspace}/bin/e2e_tests
```
dpc, line 62: **thought (non-blocking):** Oh. Interesting. I think it works, and nothing wrong with it but generally [build a whole workspace for testing and other things, and then run the e2e script](https://github.com/fedimint/fedimint/blob/44faa9a9a1945914b1f6241d0f905a854235c81b/nix/craneBuild.nix#L140).

The advantage is that the test scripts work all the same both from dev env (`./script/run-some-test.sh`) and CI.

In Flakebox I even added specialized commands to make it cleaner.

lelgenio: Certainly! I added a comment to highlight this.

lelgenio:
I went ahead and added PostgreSQL to the example, just to hammer the point home that you can spin up anything together in the test.

lelgenio:
I didn't squash so it's easy to drop postgres if necessary

examples/end-to-end-testing/flake.nix
```diff
@@ -28,10 +28,12 @@
           pname = "example-e2e";
           version = "0.1";
           doCheck = false;
-          nativeBuildInputs = lib.optionals pkgs.stdenv.isDarwin [
+          nativeBuildInputs = lib.optionals pkgs.stdenv.isDarwin (with pkgs.darwin.apple_sdk.frameworks; [
             pkgs.libiconv
-            pkgs.darwin.apple_sdk.frameworks.Security
-          ];
+            CoreFoundation
```
dpc: **question (non-blocking):** Do you know what are these for and what are a good defaults? AFAIR I always have to add `libiconv` and `Security`, but I don't even know why and what is the most commonly used set.

examples/end-to-end-testing/flake.lock
```diff
@@ -0,0 +1,128 @@
+{
+  "nodes": {
```
ipetkov, line 2: We can omit this lockfile, none of the other examples do (I'm not entirely sure if Nix will blindly copy it when used as a template or not)

flake.nix
```diff
@@ -82,6 +82,11 @@
           description = "Build a workspace with a trunk member";
           path = ./examples/trunk-workspace;
         };
+
+        end-to-end-testing = {
```
ipetkov, line 86: Nit: These are generally sorted alphabetically if you don't mind moving this up a tad!

examples/end-to-end-testing/flake.nix
ipetkov: I think darwin builds usually complain if you don't have them (I always try without them and add them in if the build fails :man_shrugging: )

ipetkov (approved):
Looking great, thanks for adding this @lelgenio !

I don't mind having postgres in there as long as it doesn't end up making the test suite run for too long (though we can always comment it out later if it does).

Few minor nit/cleanup comments and happy to merge afterwards!

examples/end-to-end-testing/flake.lock
lelgenio: Oops, how did that end up in here?

> I'm not entirely sure if Nix will blindly copy it when used as a template or not

Had to check, Yep the lockfile gets copied.

flake.nix
lelgenio: Done. I let `default` stay together with `quick-start` though, as formatting seems to imply `default` is not an actual item on the list.

examples/end-to-end-testing/flake.nix
lelgenio: I figured it out from the [logs](https://github.com/ipetkov/crane/actions/runs/6490015759/job/17625187508#step:5:2581), no idea what exactly it does.

image

PR #378, opened 2023-08-31
Rust shell

### LINKED ISSUE #377: Provide a shell derivation
The current examples recommend `pkgs.mkShell` for providing development shells:

```nix
devShells.default = pkgs.mkShell {
  inputsFrom = builtins.attrValues self.checks.${system};

  # Additional dev-shell environment variables can be set directly
  # MY_CUSTOM_DEVELOPMENT_VAR = "something else";

  # Extra inputs can be added here
  nativeBuildInputs = with pkgs; [
    cargo
    rustc
  ];
};
```

It would be nice if we could provide a `craneLib.rustShell` function that looked something like this:

```nix
{ lib
, mkShell
, rustPlatform
, rust-analyzer
, cargo
, rustc
}:

{ checks ? {}
, inputsFrom ? []
, nativeBuildInputs ? []
, rust-analyzer ? rust-analyzer
, ...
}@args:
let
  inherit (builtins) removeAttrs;

  cleanedArgs = removeAttrs args [
    "checks"
    "inputsFrom"
    "nativeBuildInputs"
    "rust-analyzer"
  ];
in
mkShell (cleanedArgs // {
  inputsFrom = builtins.attrValues checks ++ inputsFrom;

  nativeBuildInputs =
    [
      rustc
      cargo
    ]
    ++ lib.optional (rust-analyzer != null) rust-analyzer
    ++ nativeBuildInputs;
}
// lib.optionalAttrs (rust-analyzer != null) {
  RUST_SRC_PATH = rustPlatform.rustLibSrc;
})
```

Does this sound like a good idea? If so, I can put up a PR.

9999years:
## Motivation
Closes #377.

## Checklist
- [x] updated `CHANGELOG.md`

### DIFF
--- a/examples/alt-registry/flake.nix
+++ b/examples/alt-registry/flake.nix
@@ -100,7 +100,7 @@
 
           # Extra inputs can be added here; cargo and rustc are provided by default.
           packages = [
-            pkgs.ripgrep
+            # pkgs.ripgrep
           ];
         };
       });
--- a/examples/alt-registry/flake.nix
+++ b/examples/alt-registry/flake.nix
@@ -91,13 +91,16 @@
           drv = my-crate;
         };
 
-        devShells.default = pkgs.mkShell {
-          inputsFrom = builtins.attrValues self.checks.${system};
+        devShells.default = craneLib.devShell {
+          # Inherit inputs from checks.
+          checks = self.checks.${system};
 
-          # Extra inputs can be added here
-          nativeBuildInputs = with pkgs; [
-            cargo
-            rustc
+          # Additional dev-shell environment variables can be set directly
+          # MY_CUSTOM_DEVELOPMENT_VAR = "something else";
+
+          # Extra inputs can be added here; cargo and rustc are provided by default.
+          packages = [
+            pkgs.ripgrep
           ];
         };
       });
--- a/examples/build-std/flake.nix
+++ b/examples/build-std/flake.nix
@@ -82,9 +82,10 @@
           # Inherit inputs from checks.
           checks = self.checks.${system};
 
-          # Extra inputs can be added here; cargo and rustc are provided by default.
+          # Extra inputs can be added here; cargo and rustc are provided by default
+          # from the toolchain that was specified earlier.
           packages = [
-            pkgs.rustToolchain
+            # rustToolchain
           ];
         };
       });
--- a/examples/build-std/flake.nix
+++ b/examples/build-std/flake.nix
@@ -78,12 +78,13 @@
 
         packages.default = my-crate;
 
-        devShells.default = pkgs.mkShell {
-          inputsFrom = builtins.attrValues self.checks.${system};
+        devShells.default = craneLib.devShell {
+          # Inherit inputs from checks.
+          checks = self.checks.${system};
 
-          # Extra inputs can be added here
-          nativeBuildInputs = with pkgs; [
-            rustToolchain
+          # Extra inputs can be added here; cargo and rustc are provided by default.
+          packages = [
+            pkgs.rustToolchain
           ];
         };
       });
--- a/examples/custom-toolchain/flake.nix
+++ b/examples/custom-toolchain/flake.nix
@@ -68,12 +68,13 @@
           '';
         };
 
-        devShells.default = pkgs.mkShell {
-          inputsFrom = builtins.attrValues self.checks.${system};
+        devShells.default = craneLib.devShell {
+          # Inherit inputs from checks.
+          checks = self.checks.${system};
 
-          # Extra inputs can be added here
-          nativeBuildInputs = with pkgs; [
-            rustWithWasiTarget
+          # Extra inputs can be added here; cargo and rustc are provided by default.
+          packages = [
+            pkgs.rustWithWasiTarget
           ];
         };
       });
--- a/examples/custom-toolchain/flake.nix
+++ b/examples/custom-toolchain/flake.nix
@@ -72,9 +72,10 @@
           # Inherit inputs from checks.
           checks = self.checks.${system};
 
-          # Extra inputs can be added here; cargo and rustc are provided by default.
+          # Extra inputs can be added here; cargo and rustc are provided by default
+          # from the toolchain that was specified earlier.
           packages = [
-            pkgs.rustWithWasiTarget
+            # rustWithWasiTarget
           ];
         };
       });
--- a/examples/quick-start-simple/flake.nix
+++ b/examples/quick-start-simple/flake.nix
@@ -45,16 +45,16 @@
           drv = my-crate;
         };
 
-        devShells.default = pkgs.mkShell {
-          inputsFrom = builtins.attrValues self.checks.${system};
+        devShells.default = craneLib.devShell {
+          # Inherit inputs from checks.
+          checks = self.checks.${system};
 
           # Additional dev-shell environment variables can be set directly
           # MY_CUSTOM_DEVELOPMENT_VAR = "something else";
 
-          # Extra inputs can be added here
-          nativeBuildInputs = with pkgs; [
-            cargo
-            rustc
+          # Extra inputs can be added here; cargo and rustc are provided by default.
+          packages = [
+            pkgs.ripgrep
           ];
         };
       });
--- a/examples/quick-start-simple/flake.nix
+++ b/examples/quick-start-simple/flake.nix
@@ -54,7 +54,7 @@
 
           # Extra inputs can be added here; cargo and rustc are provided by default.
           packages = [
-            pkgs.ripgrep
+            # pkgs.ripgrep
           ];
         };
       });
--- a/examples/quick-start/flake.nix
+++ b/examples/quick-start/flake.nix
@@ -133,7 +133,7 @@
 
           # Extra inputs can be added here; cargo and rustc are provided by default.
           packages = [
-            pkgs.ripgrep
+            # pkgs.ripgrep
           ];
         };
       });
--- a/examples/quick-start/flake.nix
+++ b/examples/quick-start/flake.nix
@@ -124,16 +124,16 @@
           drv = my-crate;
         };
 
-        devShells.default = pkgs.mkShell {
-          inputsFrom = builtins.attrValues self.checks.${system};
+        devShells.default = craneLib.devShell {
+          # Inherit inputs from checks.
+          checks = self.checks.${system};
 
           # Additional dev-shell environment variables can be set directly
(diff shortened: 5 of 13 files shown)


examples/alt-registry/flake.nix
```diff
@@ -91,13 +91,16 @@
...
-          # Extra inputs can be added here
-          nativeBuildInputs = with pkgs; [
-            cargo
-            rustc
+          # Additional dev-shell environment variables can be set directly
+          # MY_CUSTOM_DEVELOPMENT_VAR = "something else";
+
+          # Extra inputs can be added here; cargo and rustc are provided by default.
+          packages = [
+            pkgs.ripgrep
```
ipetkov, line 103: Nit: we can leave this commented out but still illustrate the point
```suggestion
            #pkgs.ripgrep
```

examples/build-std/flake.nix
```diff
@@ -78,12 +78,13 @@
...
+        devShells.default = craneLib.devShell {
+          # Inherit inputs from checks.
+          checks = self.checks.${system};
 
-          # Extra inputs can be added here
-          nativeBuildInputs = with pkgs; [
-            rustToolchain
+          # Extra inputs can be added here; cargo and rustc are provided by default.
+          packages = [
+            pkgs.rustToolchain
```
ipetkov, line 87:
```suggestion
            #pkgs.ripgrep
```

examples/custom-toolchain/flake.nix
```diff
@@ -68,12 +68,13 @@
...
+        devShells.default = craneLib.devShell {
+          # Inherit inputs from checks.
+          checks = self.checks.${system};
 
-          # Extra inputs can be added here
-          nativeBuildInputs = with pkgs; [
-            rustWithWasiTarget
+          # Extra inputs can be added here; cargo and rustc are provided by default.
+          packages = [
+            pkgs.rustWithWasiTarget
```
ipetkov, line 77:
```suggestion
            #pkgs.ripgrep
```

examples/build-std/flake.nix
```diff
@@ -78,12 +78,13 @@
...
-        devShells.default = pkgs.mkShell {
-          inputsFrom = builtins.attrValues self.checks.${system};
+        devShells.default = craneLib.devShell {
+          # Inherit inputs from checks.
+          checks = self.checks.${system};
 
-          # Extra inputs can be added here
-          nativeBuildInputs = with pkgs; [
-            rustToolchain
+          # Extra inputs can be added here; cargo and rustc are provided by default.
```
ipetkov, line 85:
```suggestion
          # Extra inputs can be added here; cargo and rustc are provided by default
          # from the toolchain that was specified earlier.
```

examples/custom-toolchain/flake.nix
```diff
@@ -68,12 +68,13 @@
...
-        devShells.default = pkgs.mkShell {
-          inputsFrom = builtins.attrValues self.checks.${system};
+        devShells.default = craneLib.devShell {
+          # Inherit inputs from checks.
+          checks = self.checks.${system};
 
-          # Extra inputs can be added here
-          nativeBuildInputs = with pkgs; [
-            rustWithWasiTarget
+          # Extra inputs can be added here; cargo and rustc are provided by default.
```
ipetkov, line 75:
```suggestion
          # Extra inputs can be added here; cargo and rustc are provided by default
          # from the toolchain that was specified earlier.
```
```

File: examples/quick-start-simple/flake.nix [hunk 6]
```diff
@@ -45,16 +45,16 @@
...
           # Additional dev-shell environment variables can be set directly
           # MY_CUSTOM_DEVELOPMENT_VAR = "something else";
 
-          # Extra inputs can be added here
-          nativeBuildInputs = with pkgs; [
-            cargo
-            rustc
+          # Extra inputs can be added here; cargo and rustc are provided by default.
+          packages = [
+            pkgs.ripgrep
```
ipetkov, line 57:
```suggestion
            #pkgs.ripgrep
```

examples/quick-start/flake.nix
```diff
@@ -124,16 +124,16 @@
...
           # Additional dev-shell environment variables can be set directly
           # MY_CUSTOM_DEVELOPMENT_VAR = "something else";
 
-          # Extra inputs can be added here
-          nativeBuildInputs = with pkgs; [
-            cargo
-            rustc
+          # Extra inputs can be added here; cargo and rustc are provided by default.
+          packages = [
+            pkgs.ripgrep
```
ipetkov, line 136:
```suggestion
            #pkgs.ripgrep
```

lib/devShell.nix
```diff
@@ -0,0 +1,38 @@
...
+, rustPlatform
+, cargo
+, rustc
+, pkgs
+}:
+
+{ checks ? {}
+, inputsFrom ? []
+, packages ? []
+, rust-analyzer ? pkgs.rust-analyzer
```
ipetkov, line 12: I think we should avoid using `pkgs.rust-analyzer` here (instead hoist the binding to the top-level closure) so that we can pull the "correctly overlayed" version via `callPackage`.

(I really don't think anyone should be doing crazy stuff like `craneLib.overrideScope' (final: prev: { rust-analyzer = whatever; })` but in case there's shenanigans we won't make things worse)

I also anticipate it might have been written this way to avoid binding `rust-analyzer` to itself (even if the language supported it, it would look weird...). One suggestion I might provide is to not bind `rust-analyzer` here at all, but access it via another temporary variable:

```nix
let
  chosen-rust-analyzer = args.rust-analyzer or rust-analyzer;
in
lib.optional (chosen-rust-analyzer != null) chosen-rust-analyzer
```

lib/devShell.nix
```diff
@@ -0,0 +1,38 @@
...
+  packages =
+    [
+      rustc
+      cargo
+    ]
+    ++ lib.optional (rust-analyzer != null) rust-analyzer
+    ++ packages;
+}
+// lib.optionalAttrs (rust-analyzer != null) {
+  RUST_SRC_PATH = rustPlatform.rustLibSrc;
```
ipetkov, line 37: Unfortunately we can't use `rustPlatform` here since it will not know about any toolchains overlayed during instantiation (e.g. `craneLib.overrideToolchain`). 

I'm not sure of the best way to get this. I believe that any `rustc` derivations (e.g. via `oxalica/rust-overlay`) which contain the `src` component _should_ make it available somewhere, but that starts making assumptions about how toolchains are being installed/configured.

If there's another decent heuristic we can use I'd be happy to use it, but if we can't find one, I'd rather we leave this part up to whomever needs it (though we can also add it back in if we find such a heuristic later)

ipetkov (approved):
Looks great, thanks for polishing all the docs and the changelog :heart_eyes_cat: 

Just have some minor nitpicks/clean up comments, then happy to merge once CI is green

lib/devShell.nix
9999years: Hmm, I see. This variable is necessary to get `rust-analyzer` working, let me think about this.

lib/devShell.nix
9999years: > I really don't think anyone should be doing crazy stuff like `craneLib.overrideScope' (final: prev: { rust-analyzer = whatever; })`

Doesn't seem _that_ far-fetched, `rust-analzyer` moves quickly enough that people might want to get it from an overlay or a separate flake or something.

9999years: I hoisted the definition to the top-level binding — is that sufficient for getting this to work with overlays and overrides?

lib/devShell.nix
9999years: So let me check if I understand this correctly:

The issue is that `pkgs.rustPlatform` may not be the correct `rustPlatform` given `craneLib.overrideToolchain` can be used to select a different Rust platform, and the different Rust platform may not include a `rustPlatform` or `rustLibSrc` attribute?

9999years: For the `rustWithWasiTarget` toolchain/override here, the path `${rustWithWasiTarget.availableComponents.rust-src}/lib/rustlib/src/rust/library/` is equivalent to `rustPlatform.rustLibSrc`:

https://github.com/ipetkov/crane/blob/6aaaa996c4e039b85dafc1ed9d0f4f075c49b50a/examples/custom-toolchain/flake.nix#L31-L39

Maybe we could check if there's a `rustc.availableComponents` attribute and use the `oxalica/rust-overlay` source in that case, or `rustPlatform.rustLibSrc` otherwise? I'm not sure it's bulletproof though, what do you think?

9999years: Alternatively, we could make `rustLibSrc` an explicit argument to `devShell` and only set `RUST_SRC_PATH` if it's given.

ipetkov: > The issue is that `pkgs.rustPlatform` may not be the correct `rustPlatform` given `craneLib.overrideToolchain` can be used to select a different Rust platform, and the different Rust platform may not include a `rustPlatform` or `rustLibSrc` attribute?

Yep this is correct. The only time `pkgs.rustPlatform` would change is if someone overlayed a different `rustc` or `cargo` version into `nixpkgs`, but `crane`'s view of `rustc` and `cargo` can be influenced independently (via `.overrideToolchain`).

We have no guarantees about what the `rustc`/`cargo` values passed into `.overrideToolchain` are (besides the presumption that they contain their own respective binary which behaves as one might generally expect)

> For the rustWithWasiTarget toolchain/override here, the path ${rustWithWasiTarget.availableComponents.rust-src}/lib/rustlib/src/rust/library/ is equivalent to rustPlatform.rustLibSrc:

They are equivalent, yes, but not identical, meaning they can come from two entirely different toolchains (notably nixpkgs usually lags around one stable version behind Rust releases)

---

Given these challenges, for the time being (this PR) I would prefer we _neither include `rust-analyer` nor set `RUSTC_SRC_PATH`_: if we cannot get things exactly right (e.g. point `rust-analyer` to the "correct" source for the toolchain) we should not make an incorrect guess and give someone a debugging headache.

Totally happy to revisit this in the future; if the API would need to change to support it better we can also explore that in a future PR since I think there's a lot of value what's in here already!

9999years: Done!

PgBiel:
Just wanted to give some input as a fellow `crane` user: ensuring `clippy` and `rustfmt` are present (and overridable) can also prove itself to be helpful for a dev shell. (Not sure how easy that would be considering the `rust-analyzer` problem, just wanted to throw the idea around.)

ipetkov:
@PgBiel I think that's a good suggestion, though I'll open a follow up issue on it!

PgBiel:
No problem in that. Thanks!

PR #379, opened 2023-09-01
Don't install `target/release` in `cargoDoc`

9999years:
## Motivation
Right now, `craneLib.cargoDoc` output includes the contents of `target/release`:

```ShellSession
$ exa --tree --level 2 result/target/release
result/target/release
├── build
│  ├── simple-277affac19bb59bc
│  └── simple-b13f3c7ff4e09d62
├── crane-dummy -> /nix/store/sbxa6xwi2f67a2kcngcvlvsp49xicq2m-simple-deps-0.1.0/target/release/crane-dummy
├── crane-dummy.d -> /nix/store/sbxa6xwi2f67a2kcngcvlvsp49xicq2m-simple-deps-0.1.0/target/release/crane-dummy.d
├── deps
│  ├── byteorder-aeb8a302464a1b59.d -> /nix/store/sbxa6xwi2f67a2kcngcvlvsp49xicq2m-simple-deps-0.1.0/target/release/de
ps/byteorder-aeb8a302464a1b59.d
│  ├── byteorder-f76acca05a702bfb.d -> /nix/store/sbxa6xwi2f67a2kcngcvlvsp49xicq2m-simple-deps-0.1.0/target/release/deps/byteorder-f76acca05a702bfb.d
│  ├── crane_dummy-8abc31b05855626c.d -> /nix/store/sbxa6xwi2f67a2kcngcvlvsp49xicq2m-simple-deps-0.1.0/target/release/deps/crane_dummy-8abc31b05855626c.d
│  ├── crane_dummy-54bb6b807d416584.d -> /nix/store/sbxa6xwi2f67a2kcngcvlvsp49xicq2m-simple-deps-0.1.0/target/release/deps/crane_dummy-54bb6b807d416584.d
│  └── ...
├── examples
├── incremental
├── libsimple.d -> /nix/store/sbxa6xwi2f67a2kcngcvlvsp49xicq2m-simple-deps-0.1.0/target/release/libsimple.d
└── libsimple.rlib -> /nix/store/sbxa6xwi2f67a2kcngcvlvsp49xicq2m-simple-deps-0.1.0/target/release/libsimple.rlib
```

These are symlinks, so they don't add disk space, but they do clutter up the output and potentially increase the closure size (as well as adding a dependency on the `cargoArtifacts` derivation).

Adding `rm -rf target/release` to the `preInstall` hook _works_, but it feels a little messy. Maybe a better solution would be to modify `installCargoArtifactsHook.sh` to not symlink these files in the first place?

What do you think?

## Checklist
- [x] updated `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   time) but this may end up rejecting builds which were previously passing. To
   get the old behavior back, set `cargoExtraArgs = "";`
 * Fixed a bug when testing proc macro crates with `cargoNextest` on macO. ([#376](https://github.com/ipetkov/crane/pull/376))
+* `target/release` is no longer installed from `cargoDoc`.
 
 ## [0.13.1] - 2023-08-22
 
--- a/lib/cargoDoc.nix
+++ b/lib/cargoDoc.nix
@@ -3,16 +3,22 @@
 
 { cargoDocExtraArgs ? "--no-deps"
 , cargoExtraArgs ? "--locked"
+, preInstall ? ""
 , ...
 }@origArgs:
 let
   args = (builtins.removeAttrs origArgs [
     "cargoDocExtraArgs"
     "cargoExtraArgs"
+    "preInstall"
   ]);
 in
 mkCargoDerivation (args // {
   pnameSuffix = "-doc";
 
   buildPhaseCargoCommand = "cargoWithProfile doc ${cargoExtraArgs} ${cargoDocExtraArgs}";
+
+  preInstall = preInstall + ''
+    rm -rf target/release
+  '';
 })


ipetkov:
The way to actually opt out of his behavior (installing `target` at all) is [to set `doInstallCargoArtifacts = false`](https://github.com/ipetkov/crane/blob/174604795d316b75777e28185c3a4918bc69b399/docs/API.md#L894-L896)

Now I'd entertain the argument that `cargoDoc` should default `doInstallCargoArtifacts = false;` unless specified. I can't quite recall why the existing default was left in place, maybe for "simplicity" in case anyone wanted to chain their `cargoDoc` derivation, but thinking about it today I'd be in favor of having `cargoDoc` install fewer things by default

9999years:
My view is that `cargoDoc` should install only the built documentation (from `target/doc`) — does that make sense?

ipetkov:
Yep, should be address in https://github.com/ipetkov/crane/pull/381

ipetkov:
Ah wait I though `cargoDoc` already did install the html docs and we were installing the rest of the artifacts as well, doesn't seem like we were

9999years:
https://github.com/ipetkov/crane/pull/381 looks good, I'll close this in favor of that

PR #386, opened 2023-09-16
Much faster symlinking, safer interactions with `rustc`

### LINKED ISSUE #385: Building only dependencies fragile wrt symlinking strategy
Hi! Recently crane changed to symlinking rlib and rmeta files when reading dependencies. For the random reader, Crane generates the dependencies in one nix build, and then uses them in a later one. Effectively, you generate your cargo `/target` dir in one build step, and then you 'put it into place' in the next build step.

This has many practical advantages.

Obviously, this is awesome - way less time spent copying voluminous object files. Unfortunately, it leads to a common failure mode - if a dependency is out of date somehow and cargo decides it needs to rebuild it, this rebuild will fail - it tries to write to the target of the symlink and this fails, as it's a nix store path.

I've seen this caused for 2 reasons:
1. Dodgy build.rs. I found that the onnxruntime crate emitted a 'rerun-if-changed' on its own output - thereby causing the buildscript to be re-run every time. This is now fixed on their develop. I found another crate to emit a rerun-if-changed on a file that no longer exists.
2. A differing build environment. I've found this to occur for two reasons. Firstly, I attempted to make a dependency available in tests only (a test in my project relied on being able to shell out to procps binaries. This exploded. The more insidious case that I'm currently dealing with is that the OpenCV crate rebuilds depending on `CMAKE_PREFIX_PATH` not having changed. Somehow, it is changing between the root build and the 'clippy' build. The only difference I can tell is that clippy is added to the nativeBuildInputs. I'm wondering if clippy itself causes the cmake prefix path to change. Either way, build explodes.

This can be worked around by setting `doNotLinkInheritedArtifacts = true` in most cases, but not all. In particular, I struggled to get a build which used `cxx` to build, because cxx created absolute symlinks which were dangling post copy into the Nix store. This is now fixed on master of that project.

The takeaway for me is that this change makes Crane significantly more brittle when building Rust crates, and the error which is emitted can be rather hard to debug.

A sample error would be:

```
error: output file /private/tmp/nix-build-retina-clippy-0.0.0.drv-0/source/target/release/deps/libopencv-6a1a900b840d12b6.rmeta is not writeable -- check its permissions
```

which if you know how crane works, makes sense - but it's not very introspectable for abstract random dev joining the project. I'm trying to make it so that new engineers don't need to know how crane works in order to build their code - rather, they just need to know that it does.

I've got a few proposals for ways out:

*The opencv issue*

Here, crane is leading to a diff in CMAKE_PREFIX_PATH, between building deps and other tests. This genuinely just seems like a bug in crane - arguably Crane should be ensuring that insofar as the env can lead to cargo changes, it should be compatible between all builds. In my head, this is _likely_ being caused by Crane amending the nativeBuildInputs being passed in. I'm wondering if it might be more robust to add _all_ the relevant inputs to nativeBuildInputs every time, rather than amending to the minimal set. However, it's possible I've misdiagnosed this specific issue - the person I'm working with has not yet provided me with the CMAKE_PREFIX_PATH of the 'build deps' and 'run clippy' tests.

*The more general issues*

1. On supported filesystems, Crane could prefer the 'reflink' functionality used by the filesystem. This clones the file but points to the same data locks on disk as a copy-on-write clone of the data. Supported filesystems are xfs, btrfs, and in theory, APFS (although it does not work on my Mac machine, I'm not sure why at present - would need to build coreutils outside of a nix environment to properly validate whether this is a Nix or GNU issue). This work would likely represent checking whether some test file can be copied with `cp --reflink=always file1 file2` between the two directories, and upon success, doing `cp -r --reflink=always` as opposed to the symlinking. A further blocker to this _might_ be that Nix builds in temp directories on a different filesystem to the Nix store - I am not sure about the specifics. I believe this would likely be ok on btrfs but may prove more of an issue on xfs. In any case, an optimisation rather than a total fix. However, if you could make this work on Linux, it might prove better on Mac.
2. Use unionfs-fuse in order to mount the old target and new target as a union rather than as the root. This would probably be even better from a performance perspective as it leads to less directory traversal (of which much is likely redundant). However, fuse-unionfs is presently marked as broken on MacOS, so like 1, this would be a partial solution. That's not to say it can't be easily fixed however. It's likely that this would want to do some testing wrt linux versions. This would likely also require adding /dev/fuse to the nix allowlist. So possibly not viable: https://github.com/NixOS/nix/blob/2a52ec4e928c254338a612a6b40355512298ef38/src/libstore/build/local-derivation-goal.cc#L1782
3. Better debug information. Could grep for such an error in cargo output and print specific solution information. Perhaps putting this in the docs would be helpful. One other thing might be to have a sample output which just tests that rebuilds are 'safe' in this model
4. Modify Cargo to handle this case. Cargo could check if the file is writeable, and if not, and it's also a symlink, it could remove the symlink and overwrite that way. I will file a similar issue on Cargo, but have less faith on them being amenable to this because it's a Crane workflow. Alternatively, if the file needs rewriting, it's not quite as 'content addressible' as it should be, yes?
5. Not do this optimisation as 'too error prone', or at least make it be not the default.
6. It's possible that I'm also just missing something and e.g. the symlinks are somehow themselves being written as non-writeable.

Just spitballing here - don't have any particular agenda - just have personally found this confusing - and as the person doing most of the 'new nix stuff' at my company, this leads to others finding this way more confusing.

j-baker:
## Motivation

Fixes #385.

This MR represents a large performance improvement in practical terms in simple Crane situations.

For some unscientific numbers, I tested in the following fashion.

For each of the two versions of crane (master, this branch), and my simple closed source software project (it's 1kloc, about 170 crate deps, mostly just hyper):
1. I ran `nix flake check` to ensure that any crate downloads were done, any supporting derivations complete.
1. I then added a new crate (anyhow) and ran `nix flake check` again. This times the whole flow after a dependency is added.
1. I then changed a constant and re-ran, simulating the usual flow.

Because this project is relatively small, I would expect that this represents a 'worst case' scenario. For example, when uncompressed, it contains 500MB of dependencies, whereas another project I work on represents 3.6GB.

The results were as follows:
Before this PR:
- build with new dep: 2m15s
- build with new code: 1m19s After this PR:
- build with new dep: 1m22s
- build with new code: 32s

In addition, it is more robust to crate rebuilds.

How it works/why it's better:

1. Drop the diffing behaviour when doing symlinking. This is an explicit tradeoff - if one is doing symlinking on inheritance, we would expect any duplicate data to be in the form of symlinks, for which diffing file content is unhelpful. Given that this only helps the case where we are not symlinking on inheritance, are not archiving on install, it seems reasonable for it to be potentially slower in this case. I say potentially slower since if we have target dirs of 1GB, we are trading 2GB of reads for up to 1GB fewer writes. I'd note here that Nix store optimisation will cover for space savings. But, main argument: common case should be archival or symlinking, and we can boost the performance of the common case by removing this behaviour.
1. Instead, we build a `symlinks.tar` containing symlinks to the outputs of this derivation.
1. When inheriting, instead of traversing the tree and creating symlinks, we just extract this tar. This is great beacuse it means that on both derivation end and derivation start, we avoid forking O(num files produced by cargo build) processes. Since even small projects have thousands of files emitted (my own has 2033 output files). Effectively, GNU tar is much more optimised than the pre-existing bash script. In addition, the data written in the deps-only build is write-once, read many (hopefully). By pushing the directory traversal into the read side, we convert random IO into sequential.
1. At this point, we still have the problem where rustc may try to write to a file. We use a `RUSTC_WRAPPER` to instead write to a temporary directory, and after the command is finished we copy the artifacts from the out dir back to the target location. There is a potential (small) slowdown caused by this - I observe cargo to use rustc's stderr to kick off new builds as soon as it can, and so had to capture rustc's stdout. However, this effect is most likely very minor.

## Checklist

- [x] updated `CHANGELOG.md`

### DIFF
--- a/lib/cargoClippy.nix
+++ b/lib/cargoClippy.nix
@@ -19,5 +19,7 @@ mkCargoDerivation (args // {
 
   buildPhaseCargoCommand = "cargoWithProfile clippy ${cargoExtraArgs} ${cargoClippyExtraArgs}";
 
+  doInstallCargoArtifacts = false;
+
   nativeBuildInputs = (args.nativeBuildInputs or [ ]) ++ [ clippy ];
 })
--- a/lib/mkCargoDerivation.nix
+++ b/lib/mkCargoDerivation.nix
@@ -8,6 +8,7 @@
 , rsync
 , stdenv
 , vendorCargoDeps
+, writeShellApplication
 , zstd
 }:
 
@@ -39,12 +40,54 @@ let
     "pnameSuffix"
     "stdenv"
   ];
+
+  rustcWrapper = writeShellApplication {
+    name = "rustc-wrapper";
+    text = ''
+      set -euo pipefail
+
+      args=("$@")
+
+      temp_dir=$(mktemp -d)
+      trap 'rm -rf -- "$temp_dir"' EXIT
+
+      for i in "''${!args[@]}"; do
+        if [[ "''${args[i]}" == --out-dir=* ]]; then
+          current_out_dir="''${args[i]#--out-dir=}"
+          args[i]="--out-dir=$temp_dir"
+          break
+        elif [[ "''${args[i]}" == --out-dir ]]; then
+          current_out_dir="''${args[i+1]}"
+          args[i+1]="$temp_dir"
+          break
+        fi
+      done
+
+      if [[ -v current_out_dir ]]; then
+        stderr="$temp_dir/crane_stderr"
+        set +e
+        stdout=$("''${args[@]}" 2>"$stderr")
+        exit_code=$?
+        set -e
+        stderr_text=$(cat "$stderr")
+        rm "$stderr"
+        cp -r "$temp_dir/." "$current_out_dir"
+        echo -n "$stdout"
+        echo -n "$stderr_text" >&2
+        exit $exit_code
+      else
+        exec "$@"
+      fi
+    '';
+  };
 in
 chosenStdenv.mkDerivation (cleanedArgs // {
   inherit cargoArtifacts;
 
   pname = "${args.pname or crateName.pname}${args.pnameSuffix or ""}";
   version = args.version or crateName.version;
+
+  RUSTC_WRAPPER="${rustcWrapper}/bin/rustc-wrapper";
 
   # Controls whether cargo's `target` directory should be copied as an output
   doInstallCargoArtifacts = args.doInstallCargoArtifacts or true;
--- a/lib/setupHooks/inheritCargoArtifactsHook.sh
+++ b/lib/setupHooks/inheritCargoArtifactsHook.sh
@@ -50,10 +50,9 @@ inheritCargoArtifacts() {
       # being built (since changing `Cargo.lock` would rebuild everything anyway), which makes them
       # good candidates for symlinking (esp. since they can make up 60-70% of the artifact directory
       # on most projects). Thus we ignore them when copying all other artifacts below as we will
-      # symlink them afterwards. Note that we scope these checks to the `/deps` subdirectory; the
-      # workspace's own .rlib and .rmeta files appear one directory up (and these may require being
-      # writable depending on how the actual workspace build is being invoked, so we'll leave them
-      # alone).
+      # symlink them afterwards. Note that we scope these checks to the `/deps` subdirectory, as
+      # this directory contains only rustc output. We have a rustc wrapper which ensures that any
+      # writes to these symlinks do not try to change the (immutable) symlink targets.
       #
       # NB: keep the executable bit only if set on the original file
       # but make all files writable as sometimes read-only files will make the build choke
@@ -66,27 +65,15 @@ inheritCargoArtifacts() {
         --times \
         --chmod=u+w \
         --executability \
-        --exclude 'deps/*.rlib' \
-        --exclude 'deps/*.rmeta' \
+        --exclude '*/deps/*' \
         --exclude '.cargo-lock' \
         "${preparedArtifacts}/" \
         "${cargoTargetDir}/"
-
-      local linkCandidates=$(mktemp linkCandidatesXXXX.txt)
-      find "${preparedArtifacts}" \
-        '(' -path '*/deps/*.rlib' -or -path '*/deps/*.rmeta' ')' \
-        -printf "%P\n" \
-        >"${linkCandidates}"
-
-      # Next create any missing directories up front so we can avoid redundant checks later
-      cat "${linkCandidates}" \
-        | xargs --no-run-if-empty -n1 dirname \
-        | sort -u \
-        | (cd "${cargoTargetDir}"; xargs --no-run-if-empty mkdir -p)
-
-      # Lastly do the actual symlinking
-      cat "${linkCandidates}" \
-        | xargs -P ${NIX_BUILD_CORES} -I '##{}##' ln -s "${preparedArtifacts}/##{}##" "${cargoTargetDir}/##{}##"
+      
+      symlinks="${cargoArtifacts}/symlinks.tar"
+      if [[ -f "${symlinks}" ]]; then
+        tar -xf "${symlinks}" -C "${cargoTargetDir}" --strip-components=1
+      fi
     fi
   else
     echo unable to copy cargo artifacts, \"${preparedArtifacts}\" looks invalid
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,8 +18,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   `Cargo.lock` is exactly what is expected (without implicit changes at build
   time) but this may end up rejecting builds which were previously passing. To
   get the old behavior back, set `cargoExtraArgs = "";`
-* **Breaking**: `cargoDoc` will no longer install cargo artifacts by default.
-  Set `doInstallCargoArtifacts = true;` to get the old behavior back.
+* **Breaking**: `cargoDoc` and `cargoClippy` will no longer install cargo artifacts
+  by default. Set `doInstallCargoArtifacts = true;` to get the old behavior back.
+* **Breaking**: rustc is now executed with a wrapper script. Therefore providing
+  your own value of `RUSTC_WRAPPER` will no longer work. We do not believe that
+  this is commonly used.
+* Large performance improvement when symlinking artifacts.
 * `cargoDoc` will now install generated documentation in `$out/share/doc`
 * Fixed a bug when testing proc macro crates with `cargoNextest` on macOS.
   ([#376](https://github.com/ipetkov/crane/pull/376))
--- a/docs/API.md
+++ b/docs/API.md
@@ -1439,16 +1439,15 @@ identical files against a directory of previously prepared cargo artifacts.
 It takes three positional arguments:
 1. the installation directory for the output.
    * An error will be raised if not specified
-   * If the specified path is a directory which exists then the current cargo
-     artifacts will be compared with the contents of said directory. Any files
-     whose contents and paths match will be symbolically linked together to
-     reduce the size of the data stored in the Nix store.
 1. the path to cargo's artifact directory
    * An error will be raised if not specified
 1. a path to the previously prepared cargo artifacts
    * An error will be raised if not specified
    * `/dev/null` can be specified here if there is no previous directory to
      deduplicate against
+
+'symlinks.tar' will be created in the output directory in this case. It contains
+symlinks to all of the dep files in this output, simplifying the untar behaviour.
 
 Defines `prepareAndInstallCargoArtifactsDir()` which handles installing cargo's
 artifact directory to the derivation's output. It takes three positional
--- a/lib/setupHooks/installCargoArtifactsHook.sh
+++ b/lib/setupHooks/installCargoArtifactsHook.sh
@@ -25,23 +25,16 @@ dedupAndInstallCargoArtifactsDir() {
 
   mkdir -p "${dest}"
 
-  if [ -d "${prevCargoTargetDir}" ]; then
-    echo "symlinking duplicates in ${cargoTargetDir} to ${prevCargoTargetDir}"
-
-    while read -r fullTargetFile; do
-      # Strip the common prefix of the current target directory
-      local targetFile="${fullTargetFile#"${cargoTargetDir}"}"
-      # Join the path and ensure we don't have a duplicate `/` separator
-      local candidateOrigFile="${prevCargoTargetDir}/${targetFile#/}"
-
-      if cmp --silent "${candidateOrigFile}" "${fullTargetFile}"; then
-        ln --symbolic --force --logical "${candidateOrigFile}" "${fullTargetFile}"
-      fi
-    done < <(find "${cargoTargetDir}" -type f)
-  fi
-
   echo installing "${cargoTargetDir}" to "${dest}"
   mv "${cargoTargetDir}" --target-directory="${dest}"
+
+  local symlinksDir="$(mktemp -d)"
+  cp -Rs "${dir}/target" "${symlinksDir}"
+  if test -n "$(shopt -s nullglob; echo $symlinksDir/target/*/deps)"; then
+    pushd "$symlinksDir"
+    tar -cf "${dir}/symlinks.tar" target/*/deps
+    popd
+  fi
 }
 
 prepareAndInstallCargoArtifactsDir() {


lib/mkCargoDerivation.nix
```diff
@@ -39,13 +40,55 @@ let
...
+    '';
+  };
 in
 chosenStdenv.mkDerivation (cleanedArgs // {
   inherit cargoArtifacts;
 
   pname = "${args.pname or crateName.pname}${args.pnameSuffix or ""}";
   version = args.version or crateName.version;
 
+  RUSTC_WRAPPER="${rustcWrapper}/bin/rustc-wrapper";
```
dpc, line 90: I didn't even know crane has functionality like this and I was using my wrapper myself anyway. It seems to me like an unnecessary functionality to expose to the user.

dpc: While OK for PoC use, I think a proper wrapper should be handled by crane internally:

* check where the `rustc` is (with `which rustc`)
* set some `CRANE_RUSTC_ORIG_BIN` to it
* prepend `PATH` with a path to a wrapper binary
* inside the wrapper binary `exec $CRANE_RUSTC_ORIG_BIN $@ ...`

lib/mkCargoDerivation.nix
```diff
@@ -39,13 +40,55 @@ let
...
+          break
+        elif [[ "''${args[i]}" == --out-dir ]]; then
+          current_out_dir="''${args[i+1]}"
+          args[i+1]="$temp_dir"
+          break
+        fi
+      done
+
+      if [[ -v current_out_dir ]]; then
+        stderr="$temp_dir/crane_stderr"
```
dpc, line 67: Why bother with capturing stderr and stdout?

lib/mkCargoDerivation.nix
```diff
@@ -39,13 +40,55 @@ let
...
+        set +e
+        stdout=$("''${args[@]}" 2>"$stderr")
+        exit_code=$?
+        set -e
+        stderr_text=$(cat "$stderr")
+        rm "$stderr"
+        cp -r "$temp_dir/." "$current_out_dir"
```
dpc, line 74: `cp -a`?

lib/mkCargoDerivation.nix
```diff
@@ -39,13 +40,55 @@ let
...
+        fi
+      done
+
+      if [[ -v current_out_dir ]]; then
+        stderr="$temp_dir/crane_stderr"
+        set +e
+        stdout=$("''${args[@]}" 2>"$stderr")
+        exit_code=$?
+        set -e
+        stderr_text=$(cat "$stderr")
```
dpc, line 72: Just a thought - if you know out dir ... wouldn't it be faster & easier to delete it beforehand instead of changing it and copying from it? Technically cargo might want to read something from it first... but something tells me it never actually happen.

dpc: I would also be a bit worries about these temp paths being written somewhere and then causing rebuilds etc. Instead of making them random, maybe they should be fixed. Eg. always `$TMPDIR/rustc-wrapper/${orig-outdir-absolute-path}`.

lib/setupHooks/inheritCargoArtifactsHook.sh
```diff
@@ -66,27 +65,15 @@ inheritCargoArtifacts() {
...
-        | sort -u \
-        | (cd "${cargoTargetDir}"; xargs --no-run-if-empty mkdir -p)
-
-      # Lastly do the actual symlinking
-      cat "${linkCandidates}" \
-        | xargs -P ${NIX_BUILD_CORES} -I '##{}##' ln -s "${preparedArtifacts}/##{}##" "${cargoTargetDir}/##{}##"
+      
+      symlinks="${cargoArtifacts}/symlinks.tar"
+      if [[ -f "${symlinks}" ]]; then
+        tar -xf "${symlinks}" -C "${cargoTargetDir}" --strip-components=1
```
dpc, line 75: Would still compress.

dpc: Question: Now that everything is a symlink, but the symlink is invisible to Nix because it's in a tar archive, what is going to prevent Nix from GCing the step being linked to?

lib/mkCargoDerivation.nix
dpc: Oh. Cargo uses it to know when a process finished, instead of waiting for it to fully terminate?

:/

dpc: Thought: `rm -Rf out-dir` before starting original, would take care of this, get rid of an EXIT hook, temp dir altogether. If it worked.

But it just occurred to me that it's `out-DIR` and probably oftentimes the same dir is used to write many files, so that probably doesn't fly.

dpc: One could use `LD_PRELOAD` to intercept `open`, `write` syscalls in a rustc-wrapper and basically always follow the link if file is being read, and delete the symlink if being written to.

https://osterlund.xyz/posts/2018-03-12-interceptiong-functions-c.html

A bit involved, but could probably work perfectly and be 100% invisible.

dpc:
I think this approach can work. It's not entirely invisible, and I raised some concerns, but probably it will, at least most of the time.


> * build with new dep: 2m15s
> * build with new code: 1m19s
> After this PR:
> * build with new dep: 1m22s
> * build with new code: 32s

Would love to see how zst mode fares for comparison.

dpc:
@j-baker Ok, so I tried it in https://github.com/fedimint/fedimint/pull/3222 and it fails with:

```
fedimint-workspace-build>    Compiling fedimint-dbtool v0.1.0 (/build/source/fedimint-dbtool)
fedimint-workspace-build> cp: cannot create regular file '/build/source/target/ci/deps/./fedimint_wasm_tests.d': Permission denied
```

Test suite here also fails, so something is off somewhere. I'm happy to try again future iterations.

lib/cargoClippy.nix
```diff
@@ -19,5 +19,7 @@ mkCargoDerivation (args // {
 
   buildPhaseCargoCommand = "cargoWithProfile clippy ${cargoExtraArgs} ${cargoClippyExtraArgs}";
 
+  doInstallCargoArtifacts = false;
```
ipetkov, line 22: Why was this change needed?

I don't agree with changing the defaults here because `cargoClippy` is intended to be allow chaining with other builds by default

lib/mkCargoDerivation.nix
ipetkov, line 90: I'm strongly opposed to the forced usage of `RUSTC_WRAPPER` here. Just because _we believe_ it to be rarely used is not sufficient to _outright preclude_ projects from using their own `RUSTC_WRAPPER`.

(not to mention that _plenty_ of builds work just fine without needing this workaround, so it's the exceptional builds which should opt into different behavior than forcing it on everyone)

lib/setupHooks/inheritCargoArtifactsHook.sh
ipetkov: @dpc's observation here is correct. Nix will perform a naive scan of the results for any store paths, but will not notice them if they are behind a tarball, meaning a store GC would end up invalidating references.

ipetkov (commented):
Hi @j-baker thanks for the PR!

My main hesitation with this approach is forcing our own `RUSTC_WRAPPER` definition. I'm not sure it's even feasible to try to respect a project's existing `RUSTC_WRAPPER` (it can be set both by env var or in the cargo configuration, and I would like to avoid having to to implement cargo's lookups). Even making this opt-in is still tricky having to maintain two different implementations...

It seems like the main performance issue is doing a (single-threaded) file comparison to figure out what to symlink, but [this has given me some inspiration](https://github.com/ipetkov/crane/pull/387#discussion_r1328199880) that it may be possible to do much more cheaply with a `find` invocation which checks for updated timestamps which would be _way_ faster than actually reading out all files!

I'll try to explore that direction sometime soon, but feel free to beat me to it! My hope is that it would give us similar performance benefits with a simpler implementation :smile:

lib/setupHooks/inheritCargoArtifactsHook.sh
j-baker: An uncompressed tar file is just the uncompressed data with some special headers - no fancy behaviour. Likewise, the content of a symlink is just the path. My understanding is that Nix will have no trouble at all understanding the dependency. Hex editor image placed below.

image

j-baker: However, when compressed, I'd expect it's hard for nix to find what it's looking for, hence the explicit choice to not compress.

lib/mkCargoDerivation.nix
j-baker: Yeah, I considered this. The problem is that what rustc outputs is dependent on what rustc has been asked to do (tell it to make a dylib? it'll make one!), so it's much more tightly coupled. Whereas the output dir is loosely coupled.

Regarding the temp paths being written somewhere - I logicked myself that this should never happen. In particular, compilers put a lot of energy into being reproducible - it's dodgy if the output is dependent on the path the output is to be written to.

lib/mkCargoDerivation.nix
j-baker: There's no good ld_preload equivalent on macos which is why i didn't go down that path - and that seems extremely brittle to recompiles. Ideally Nix would let us mount fuse filesystems inside a build - that'd be the 'reliable' solution to this.

I felt this traded off nicely between being robust to change but still getting the job done.

lib/mkCargoDerivation.nix
j-baker: ack - I can handle the wrapper internally (and then also wrap with the user's wrapper).

lib/setupHooks/inheritCargoArtifactsHook.sh
dpc: That might work, but I wonder: would compressing + putting 1 symlink somewhere in the output (outside of what is copied to `./target`), keep the GC happy? Aren't all these paths pointing to the same things in nix store (or the thing it points to)?

lib/cargoClippy.nix
j-baker: I observed this to be a large slowdown in practice - specifically around copying and patching interpreter paths in the output. My intuition was that folks are using cargoClippy as a linting task, and so like cargoDoc in the last version, it was an oversight which would primarily affect new users.

What would cargoClippy chaining look like for you? Is there an example?

lib/setupHooks/inheritCargoArtifactsHook.sh
j-baker: it'd definitely work - mostly was going for simplicity vs optimality. A 345MB deps dir on my machine ends up with a 900kB symlinks.tar. Might be more clean to explicitly add a symlink though from the perspective of making the purpose clear.

j-baker:
@ipetkov for the overall rustc_wrapper problem - my problem is that right now, any time that rustc tries to overwrite a pre-existing symlink, it crashes the build (see the linked issue) - and this problem can easily be triggered by a poorly written crate (e.g. emitting a 'rerun-if-changed' declaration pointing to a path that doesn't exist). The effect is that on approx 1/2 of the projects I've migrated to Crane, I have had to patch an upstream. My sense is that the only fixes that really help this are changes in rustc (now i know more, not cargo) (which will face an uphill battle due to the difficulties of specifying user intentions on the target dir and nicheness of the workflow), a rustc wrapper, or somehow making the symlinks not be symlinks. @dpc's focus is on the latter, but this was an attempt at ensuring that symlinks degrade nicely in the case of bad crates, rather than exploding in your face with a weird error.

I can certainly write something which wraps up the user's RUSTC_WRAPPER if provided rather than removing their ability to specify one. Regarding the config.toml setting of this - doesn't crane replace the user's config.toml in the vendorCargoDeps step? That's why I didn't consider that case.

Do you have any ideas?

lib/cargoClippy.nix
dpc: While I agree that cargo clippy should be a a leaf node 99% of the time, if someone has a good reason to run it other way it should work. It technically can produce artifacts (same as `cargo check` iirc).

If I read the code here right this change doesn't just change the default, but hardcodes `false`.

Probably best as a separate PR with changelog info about change in behavior.

lib/mkCargoDerivation.nix
dpc: Yeah, I don't see a better way.

lib/setupHooks/inheritCargoArtifactsHook.sh
ipetkov: As long as there is a single line somewhere in the output (which can be a separate text file and not necessarily the tarball) which includes `/nix/store/...-blah` then Nix will consider that path a dependency of the derivation. So it's okay (as far as the GC is concerned) to create a tarball of symlinks, so long as there is some other symlink or text entry which points to the same "parent" directory that the tarball's symlinks will end up pointing to

lib/cargoClippy.nix
ipetkov: Yep if you'd like to propose a different default value here feel free to submit a separate PR for it (then we can at least have a more focused discussion on it!)

lib/mkCargoDerivation.nix
ipetkov: I'm still concerned that this would be too brittle to maintain at the library level (you can set a rustc-wrapper in `.cargo/config.toml`, and those can be littered in different places in the project).

ipetkov:
@j-baker 

> Regarding the config.toml setting of this - doesn't crane replace the user's config.toml in the vendorCargoDeps step?

No, we never blow that away outright, we simply put our own configuration "higher-up" the hierarchy (basically at the equivalent of `$HOME/.cargo`) which cargo will honor _along with_ the project's actual configuration. (I've seen projects which will dynamically update/patch their own `.cargo/config.toml` depending on what kind of builds they are doing, and so `crane` explicitly stays away from making an impact there)

> The effect is that on approx 1/2 of the projects I've migrated to Crane, I have had to patch an upstream. 

Do you have any (ideally minimal) reproductions I can look at? Ultimately for whatever reason, cargo/rustc is becoming convinced that a file needs to be rebuilt and trying to update it in place, but to me this smells more like a cache invalidation bug and maybe fixing that side steps the issue entirely...

dpc:
As for wrapping. It is possible to make a fully invisible and reliable wrapper for any binary, as long you have the path for the original. Example of the wrapper from our codebase: https://github.com/fedimint/fedimint/blob/6c15b0ea133b5239085884c14c334623287d555b/flake.toolchain.nix#L130 .

As crane is always in control of the commands that are being executed it can make something even simpler:

```bash
export RUSTC_ORIGINAL_BINARY="$(which rustc)"
export PATH="/path/to/the/wrapper-dir/:$PATH"
```

where `/path/to/the/wrapper-dir/rustc` does it thing and forward calls to the original via `exec $RUSTC_ORIGINAL_BINARY "$@"`

ipetkov:
@dpc that's not 100% reliable as one could set `build.rustc-wrapper = "/path/to/my/not-rustc"`. It's a losing battle to guess what esoteric things projects might be doing

dpc:
@ipetkov Oh, so it's a rust (mis-)feature? Terrible. I think it would be fine to just not support it. Especially if there are other crane modes which are fine with it. "If you use this feature, you need use a different mode".

ipetkov:
I wouldn't call it a misfeature, it's just not designed for "transparent" pass through wrappers

The philosophy of crane is to be a shim between cargo and Nix but make as few impacts on project decisions

dpc:
> I wouldn't call it a misfeature, it's just not designed for "transparent" pass through wrappers

It's a one-off complication for something that can be elegantly solved by a generic solution of PATH lookup order. Probably it's there because without Nix transparently wrapping a binary is actually a bad UX.

> The philosophy of crane is to be a shim between cargo and Nix but make as few impacts on project decisions

I get it, I think it's a worthy goal, but if supporting `RUSTC_WRAPPER` is what is blocking a big improvements, I think it's fine to not support.

Alternatively this PR should just implement another mode, with previous mode retained. [As I commented in `incremental zstd" PR](https://github.com/ipetkov/crane/pull/387#discussion_r1328235094), I think just maintaining a handful of modes is the way to go. You can default to the "most compatible" one, and let users switch in and out of more specialized ones, depending on their situation (which might be changing as their project evolves). Obviously too many modes is undesireable, but a handful should be OK, especially if they're different in some minor ways (like optimizations that can break in certain conditions).

j-baker:
@ipetkov an example of a crate with which I had trouble was https://github.com/microsoft/onnxruntime/pull/17018. Here, because the buildscript unnecessarily did a rerun-if-changed, rustc attempts to recompile. The crate is then unusable with the symlinks strategy, because rustc cannot overwrite a symlink which points inside the Nix store. I have more details on #385. My present situation I haven't bottomed out, but so far as I can tell, adding clippy to the nativeBuildInputs changes the CMAKE_PREFIX_PATH and this causes the `opencv` crate to decide it needs to recompile. On its own, not a problem, but it blocked a colleague for quite some time.

ipetkov:
> As you can see - I eventually got it fixed - but this took 25 days on one crate and is unreleased - in no way unreasonable for them as an OSS project, but quite a bit of time to have much slower builds.

I'm symathetic to being stuck on upstreams incorporating and releasing fixes, but ultimately buggy software is buggy and workarounds should be workarounds (IMO) instead of defaults in these situations

> The present strategy is problematic with complex buildscripts. If we have a solution which uses RUSTC_WRAPPER, then it seems relatively straightforward to instruct the user they need to change strategy if they override it.

I'd still prefer to have this be something projects can _opt-into_ rather than having to opt-out of it. If you can rework this PR so that it's opt-in (preferably by having the user override `RUSTC_WRAPPER = craneLib.whateverAppropriateNameYoudLikeToChoose ...;`) then I wouldn't mind it being available for others!

dpc:
> If you can rework this PR so that it's opt-in (preferably by having the user override `RUSTC_WRAPPER = craneLib.whateverAppropriateNameYoudLikeToChoose ...;`) then I wouldn't mind it being available for others!

How about a `mode = "use-symlink-extreme"` or something? It could be marked as experimental etc.

> but ultimately buggy software is buggy and workarounds should be workarounds (IMO) instead of defaults in these situations

I'm not sure what do you mean? Would you be fine to include it as a special opt-in mode?

Issue #370, opened 2023-08-22
Newer cargo/toolchain versions cause rebuilds of `build.rs`-using deps 


dpc:
I've recently upgraded Rust from 1.67.0 to 1.71.0, and noticed that in our "build workspace" derivation, some dependencies are being re-compiled, even though they were compiled in "build workspace deps only" derivation.

I confirmed that just switching between 1.67.0 and 1.71.0 fixes/introduces the problem.

It looks like only `build.rs` (and deps that depend on them) are being rebuilt, and from the initial debugging session I suspect something about tracking which deps require rebuild changed between these two toolchain (well... probably cargo) versions.

Notably the files restored in `/build/source/target/` from `zstd` (I use the zstd mode) archive produced by deps-only have different timestamps than originals: they are all reset to the default unix timestamp, while the original ones had the original timestamps. I suspect something about how cargo tracks these changed.

dpc:
1.68.0 is unaffected, and 1.69 is affected.

dpc:
Ironically the change was done there to display reasons for rebuilds, but the reason for rebuilding affected crates is not being displayed.

https://github.com/rust-lang/cargo/pull/11407

ipetkov:
Do you have a minimal repro I can take a look at?

What does cargo say about which files have changed specifically?

dpc:
> Do you have a minimal repro I can take a look at?

Will try to minimize when I find some time.

> What does cargo say about which files have changed specifically?

I can't get it to tell me anything, even with `-vv`. It just prints the `Compiling proc-macros2` etc., and executes the build script.

ipetkov:
You can also try running it with `CARGO_LOG=cargo::core::compiler::fingerprint=trace` and see if that shows anything interesting

dpc:
I created a quick-start example based project where I've added syn as a depdency (which brings also proc-macro2), and couldn't reprod, even bringing more and more little details of the affected project.

`CARGO_LOG` of affected vs example shows a important difference:

Affected:

```
[2023-08-23T18:58:43Z DEBUG cargo::core::compiler::fingerprint] old local fingerprints deps "/nix/store/8xlxixbffyyra5w4xv8fabyfcwynfdh0-vendor-cargo-deps/c19b7c6f923b580ac259164a89f2577984ad5ab09ee9d583b888f934adbbe8d0/proc-macro2-1.0.66" precalculated="1.0.66"
```

Not affected:

```
[...]

PR #387, opened 2023-09-17
feature: incremental zstd mode

### LINKED ISSUE #76: A way to output just a diff of `./target`

dpc:
## Motivation

Fix #76 (implement it).

This changes the `use-zstd` to create a chain of `target.tar.zst` files, each time capturing only newly created/modified files, using `mtime`, making breaking multiple `cargo` invocations into a chain of separate derivations much cheaper (virtually free in terms of disk space, extraction cost proportional to the size of whole `./target`, and compression cost relative to the size of only modified/new files).

Previous behavior (taking full snapshots of `./target` each time) is retained for time being under a new name (`use-zstd-no-incr`), as there's no guarantee that no one ever will hit a case where something modifies a file without changing mtime (though it seems unlikely). Supporting both is very easy anyway. 

**Edit**: Only later it occurred to me that incremental approach does not correct delete files (it will keep re-storing them). It could be handled, but seems unnecessary and it will have a perf cost. Current behavior is most probably OK in practice for 99.9% cases. AFAIR science goes, there are no credible reports of `./target` directory ever shrinking in the nature. But an additional reason to keep the old method around.

Possibly there should be:

* `use-zstd-incremental`
* `use-zstd-full`
* `use-zstd` (alias for `use-zstd-incremental`, or any method deemed even better in the future).

This way if anyone has a step that actually delete things, they can use `use-zstd-full` as a one-off. It might be also useful if someone really wants to introduce a break in the chain (e.g. they have really, really long chain and it seems worth it).

## Checklist

- [ ] updated `CHANGELOG.md`

### DIFF
--- a/lib/setupHooks/inheritCargoArtifactsHook.sh
+++ b/lib/setupHooks/inheritCargoArtifactsHook.sh
@@ -17,10 +17,18 @@ inheritCargoArtifacts() {
 
   mkdir -p "${cargoTargetDir}"
   if [ -f "${preparedArtifacts}" ]; then
+
+    if [ -f "${preparedArtifacts}.prev" ]; then
+      inheritCargoArtifacts "$(readlink -f "${preparedArtifacts}.prev")" "$cargoTargetDir"
+    fi
+
     echo "decompressing cargo artifacts from ${preparedArtifacts} to ${cargoTargetDir}"
 
     zstd -d "${preparedArtifacts}" --stdout | \
       tar -x -C "${cargoTargetDir}" --strip-components=1
+    rm -f "${cargoTargetDir}/.crane-previous-archive"
+    ln -s "${preparedArtifacts}" "${cargoTargetDir}/.crane-previous-archive"
+
   elif [ -d "${preparedArtifacts}" ]; then
     echo "copying cargo artifacts from ${preparedArtifacts} to ${cargoTargetDir}"
 
--- a/lib/setupHooks/installCargoArtifactsHook.sh
+++ b/lib/setupHooks/installCargoArtifactsHook.sh
@@ -87,11 +87,11 @@ prepareAndInstallCargoArtifactsDir() {
   mkdir -p "${dir}"
 
   case "${mode}" in
-    "use-zstd")
+    "use-zstd"|"use-zstd-diff")
       compressAndInstallCargoArtifactsDirIncremental "${dir}" "${cargoTargetDir}"
       ;;
 
-    "use-zstd-no-incr")
+    "use-zstd-full")
       compressAndInstallCargoArtifactsDir "${dir}" "${cargoTargetDir}"
       ;;
 
--- a/lib/setupHooks/installCargoArtifactsHook.sh
+++ b/lib/setupHooks/installCargoArtifactsHook.sh
@@ -19,6 +19,38 @@ compressAndInstallCargoArtifactsDir() {
   )
 }
 
+compressAndInstallCargoArtifactsDirIncremental() {
+  local dir="${1:?destination directory not defined}"
+  local cargoTargetDir="${2:?cargoTargetDir not defined}"
+
+  mkdir -p "${dir}"
+
+  local dest="${dir}/target.tar.zst"
+  echo "compressing ${cargoTargetDir} to ${dest}"
+  (
+    export SOURCE_DATE_EPOCH=1
+    touch -d @${SOURCE_DATE_EPOCH} "${TMPDIR}/.crane.source-date-epoch"
+
+    find "${cargoTargetDir}" \
+      -newer "${TMPDIR}/.crane.source-date-epoch" \
+      -print0 \
+      | tar \
+      --null \
+      --no-recursion \
+      --sort=name \
+      --mtime="@${SOURCE_DATE_EPOCH}" \
+      --owner=0 \
+      --group=0 \
+      --numeric-owner \
+      --pax-option=exthdr.name=%d/PaxHeaders/%f,delete=atime,delete=ctime \
+      -c -f - -T - \
+      | zstd "-T${NIX_BUILD_CORES:-0}" -o "${dest}"
+     if [ -e "${cargoTargetDir}/.crane-previous-archive" ]; then
+       cp -a "${cargoTargetDir}/.crane-previous-archive" "${dest}.prev"
+     fi
+   )
+}
+
 dedupAndInstallCargoArtifactsDir() {
   local dest="${1:?destination directory not defined}"
   local cargoTargetDir="${2:?cargoTargetDir not defined}"
@@ -56,6 +88,10 @@ prepareAndInstallCargoArtifactsDir() {
 
   case "${mode}" in
     "use-zstd")
+      compressAndInstallCargoArtifactsDirIncremental "${dir}" "${cargoTargetDir}"
+      ;;
+
+    "use-zstd-no-incr")
       compressAndInstallCargoArtifactsDir "${dir}" "${cargoTargetDir}"
       ;;
 
--- a/lib/setupHooks/installCargoArtifactsHook.sh
+++ b/lib/setupHooks/installCargoArtifactsHook.sh
@@ -88,7 +88,15 @@ prepareAndInstallCargoArtifactsDir() {
 
   case "${mode}" in
     "use-zstd"|"use-zstd-diff")
-      compressAndInstallCargoArtifactsDirIncremental "${dir}" "${cargoTargetDir}"
+
+      # If no artifcats were inherited in the first place,
+      # it's both faster and safer (mtime handlings bugs), to use
+      # a full snapshot.
+      if [ -n "${cargoArtifacts}" ];then
+        compressAndInstallCargoArtifactsDir "${dir}" "${cargoTargetDir}"
+      else
+        compressAndInstallCargoArtifactsDirIncremental "${dir}" "${cargoTargetDir}"
+      fi
       ;;
 
     "use-zstd-full")
--- a/lib/setupHooks/installCargoArtifactsHook.sh
+++ b/lib/setupHooks/installCargoArtifactsHook.sh
@@ -1,11 +1,11 @@
-compressAndInstallCargoArtifactsDir() {
+compressAndInstallCargoArtifactsDirFull() {
   local dir="${1:?destination directory not defined}"
   local cargoTargetDir="${2:?cargoTargetDir not defined}"
 
   mkdir -p "${dir}"
 
   local dest="${dir}/target.tar.zst"
-  echo "compressing ${cargoTargetDir} to ${dest}"
+  echo "compressing full content of ${cargoTargetDir} to ${dest}"
   (
     export SOURCE_DATE_EPOCH=1
     tar --sort=name \
@@ -19,14 +19,14 @@ compressAndInstallCargoArtifactsDir() {
   )
 }
 
-compressAndInstallCargoArtifactsDirIncremental() {
+compressAndInstallCargoArtifactsDirDiff() {
   local dir="${1:?destination directory not defined}"
   local cargoTargetDir="${2:?cargoTargetDir not defined}"
 
   mkdir -p "${dir}"
 
   local dest="${dir}/target.tar.zst"
-  echo "compressing ${cargoTargetDir} to ${dest}"
+  echo "compressing new content of ${cargoTargetDir} to ${dest}"
   (
     export SOURCE_DATE_EPOCH=1
     touch -d @${SOURCE_DATE_EPOCH} "${TMPDIR}/.crane.source-date-epoch"
@@ -92,15 +92,15 @@ prepareAndInstallCargoArtifactsDir() {
       # If no artifcats were inherited in the first place,
       # it's both faster and safer (mtime handlings bugs), to use
       # a full snapshot.
-      if [ -n "${cargoArtifacts}" ];then
-        compressAndInstallCargoArtifactsDir "${dir}" "${cargoTargetDir}"
+      if [ -z "${cargoArtifacts}" ];then
+        compressAndInstallCargoArtifactsDirFull "${dir}" "${cargoTargetDir}"
       else
-        compressAndInstallCargoArtifactsDirIncremental "${dir}" "${cargoTargetDir}"
+        compressAndInstallCargoArtifactsDirDiff "${dir}" "${cargoTargetDir}"
       fi
       ;;
 
     "use-zstd-full")
-      compressAndInstallCargoArtifactsDir "${dir}" "${cargoTargetDir}"
+      compressAndInstallCargoArtifactsDirFull "${dir}" "${cargoTargetDir}"
       ;;
 
     "use-symlink")
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   offline evaluations. The `outputHashes` attribute can now be optionally
   specified in `vendorCargoDeps`, `vendorGitDeps`, `vendorMultipleCargoDeps`, or
   anything else which delegates to them.
+* `use-zstd` mode uses a chained, incremental approach to avoid redundancy.
+  Old behavior (taking a full snapshot of the cargo artifacts can be achieved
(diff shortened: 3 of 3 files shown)


dpc:
I can't just let @j-baker win with #386 after all that arguing. ;)

Turns out it was surprisingly easy, and in my local tests looks like it shrank the second state target archive from 12GB to 6GB (which then gets compressed to 2.08GB)


Here is the PR using for my project where I care about it most: https://github.com/fedimint/fedimint/pull/3219

Here is the last merge to master build: https://github.com/fedimint/fedimint/actions/runs/6211721161/job/16861196422

I will compile some stats soon, or tomorrow.

dpc:
All right.  There was one bug (must use `--no-recursion`). Also the 12GB I mentioned before was largely caused by the #370 , so the new baseline seems to be 6GB compressed to 2GB (yay!).

The timestamps are mixed, as I had to re-run, and for some there was no point in re-copy-pasting.

### Before

deps only, compression:

```
Sun, 17 Sep 2023 07:56:12 GMT fedimint-deps> compressing target to /nix/store/bpma92pg41jbk5h0a9w1fp5p4haxz1jg-fedimint-deps-0.0.1/target.tar.zst
Sun, 17 Sep 2023 07:56:16 GMT fedimint-deps> /*stdin*\            : 22.28%   (  3.59 GiB =>    818 MiB, /nix/store/bpma92pg41jbk5h0a9w1fp5p4haxz1jg-fedimint-deps-0.0.1/target.tar.zst)
```

workspace, decompression:

```
Sun, 17 Sep 2023 07:28:04 GMT fedimint-workspace-build> decompressing cargo artifacts from /nix/store/bpma92pg41jbk5h0a9w1fp5p4haxz1jg-fedimint-deps-0.0.1/target.tar.zst to target
Sun, 17 Sep 2023 07:28:08 GMT fedimint-workspace-build> configuring
```

workspace, compression:

```
Sun, 17 Sep 2023 07:57:11 GMT fedimint-workspace-build> compressing target to /nix/store/0kh9szbry5zhc3fvvlsh4scgs0p43cwp-fedimint-workspace-build-0.0.1/target.tar.zst
Sun, 17 Sep 2023 07:57:18 GMT fedimint-workspace-build> /*stdin*\            : 30.91%   (  6.33 GiB =>   1.96 GiB, /nix/store/0kh9szbry5zhc3fvvlsh4scgs0p43cwp-fedimint-workspace-build-0.0.1/target.tar.zst)
```


tests, decompression:

```
Sun, 17 Sep 2023 07:31:53 GMT fedimint-test-all> decompressing cargo artifacts from /nix/store/0kh9szbry5zhc3fvvlsh4scgs0p43cwp-fedimint-workspace-build-0.0.1/target.tar.zst to target
Sun, 17 Sep 2023 07:32:00 GMT fedimint-test-all> configuring
```


# After

deps only, compression:

```
Sun, 17 Sep 2023 07:32:30 GMT fedimint-deps> compressing target to /nix/store/xvqh9q9sadqzqhc4150hxkc3q9v8rlhl-fedimint-deps-0.0.1/target.tar.zst
Sun, 17 Sep 2023 07:32:34 GMT fedimint-deps> /*stdin*\            : 21.54%   (  3.74 GiB =>    825 MiB, /nix/store/xvqh9q9sadqzqhc4150hxkc3q9v8rlhl-fedimint-deps-0.0.1/target.tar.zst)
```

workspace, decompression: 

```
Sun, 17 Sep 2023 07:32:37 GMT fedimint-workspace-build> decompressing cargo artifacts from /nix/store/xvqh9q9sadqzqhc4150hxkc3q9v8rlhl-fedimint-deps-0.0.1/target.tar.zst to target
Sun, 17 Sep 2023 07:32:41 GMT fedimint-workspace-build> configuring
```

workspace, compression:

```
Sun, 17 Sep 2023 07:51:03 GMT fedimint-workspace-build> compressing target to /nix/store/by7kqaq2jip4h2687hzjhylfdgldmj9f-fedimint-workspace-build-0.0.1/target.tar.zst
Sun, 17 Sep 2023 07:51:06 GMT fedimint-workspace-build> /*stdin*\            : 42.13%   (  2.75 GiB =>   1.16 GiB, /nix/store/by7kqaq2jip4h2687hzjhylfdgldmj9f-fedimint-workspace-build-0.0.1/target.tar.zst)
```

tests, decompression:

```
Sun, 17 Sep 2023 07:52:59 GMT fedimint-test-all> decompressing cargo artifacts from /nix/store/slislj5ja57zb9jsd7vwqkf7dyh58vxn-fedimint-deps-0.0.1/target.tar.zst to target
Sun, 17 Sep 2023 07:53:02 GMT fedimint-test-all> decompressing cargo artifacts from /nix/store/by7kqaq2jip4h2687hzjhylfdgldmj9f-fedimint-workspace-build-0.0.1/target.tar.zst to target
Sun, 17 Sep 2023 07:53:05 GMT fedimint-test-all> configuring
```

# Summary,

```
                 before             vs   after
               size       ;   c;   d             size    ;   c;  d
deps-only: 3.59G =>  818 M;  4s;  4s  vs   3.74G =>  818M;  4s; 4s
workspace: 6.33G => 1.96 G;  7s;  7s  vs   2.75G => 1.16G;  3s; 6s
```

As far as I can tell it is working as expected.

lib/setupHooks/installCargoArtifactsHook.sh
```diff
@@ -55,6 +87,10 @@ prepareAndInstallCargoArtifactsDir() {
 
   case "${mode}" in
     "use-zstd")
+      compressAndInstallCargoArtifactsDirIncremental "${dir}" "${cargoTargetDir}"
+      ;;
+
+    "use-zstd-no-incr")
```
ipetkov, line 93: I appreciate your attention to allowing one to opt into the previous behavior!

I'd okay to call out the new behavior (in the change log) and add a new environment variable (e.g. `doNotCompressIncrementally`) to `installCargoArtifactsHook.sh` which would give the old behavior back!

lib/setupHooks/installCargoArtifactsHook.sh
```diff
@@ -18,6 +18,38 @@ compressAndInstallCargoArtifactsDir() {
...
+  mkdir -p "${dir}"
+
+  local dest="${dir}/target.tar.zst"
+  echo "compressing ${cargoTargetDir} to ${dest}"
+  (
+    export SOURCE_DATE_EPOCH=1
+    touch -d @${SOURCE_DATE_EPOCH} "${TMPDIR}/.crane.source-date-epoch"
+
+    find "${cargoTargetDir}" \
+      -newer "${TMPDIR}/.crane.source-date-epoch" \
```
ipetkov, line 34: Great technique! Totally did not cross my mind that `find` can do look ups using mtime, great way to quickly find out what changed instead of having to compare (or even hash) files!

lib/setupHooks/inheritCargoArtifactsHook.sh
```diff
@@ -17,10 +17,18 @@ inheritCargoArtifacts() {
...
+    if [ -f "${preparedArtifacts}.prev" ]; then
+      inheritCargoArtifacts "$(readlink -f "${preparedArtifacts}.prev")" "$cargoTargetDir"
+    fi
+
     echo "decompressing cargo artifacts from ${preparedArtifacts} to ${cargoTargetDir}"
 
     zstd -d "${preparedArtifacts}" --stdout | \
       tar -x -C "${cargoTargetDir}" --strip-components=1
+    rm -f "${cargoTargetDir}/.crane-previous-archive"
+    ln -s "${preparedArtifacts}" "${cargoTargetDir}/.crane-previous-archive"
```
ipetkov, line 30: This works fine, though technically we're wasting a few operations waiting for the file system to delete/create the new marker file. Maybe we could store this in a (global) variable to be a bit cheaper?

ipetkov (commented):
Looks great, thanks for implementing this! The solution turned out way simpler than I had feared :tada: 

Have a few minor clean up comments (plus we need to update the changelog), but happy to merge when they are addressed!

lib/setupHooks/inheritCargoArtifactsHook.sh
dpc: I did not know (still don't) how to do that. :D

dpc: Maybe we can land as is and then follow-up with improvement?

BTW. I'm totally fine if you want to just redo/tweak and merge this PR as you please. For final touches it is generally faster than going back and forth.  :)

lib/setupHooks/installCargoArtifactsHook.sh
dpc: That $20/mo for chatgpt is paying for itself... :D

lib/setupHooks/inheritCargoArtifactsHook.sh
ipetkov: No problem, I can take it to the finish line sometime this week!

lib/setupHooks/installCargoArtifactsHook.sh
dpc: I'd call it `useFullCargoArtifactsSnapshot`. I think in certain cases it might be useful to use it even when there are no problems with the incremental approach (just to shed the chain of incremental archives, etc.)

However I'm confused how to propagate it to this function. Can I just use a global variable?

BTW. When I starting documentation changes, it did occurred to me again that I think it's better to just use different `mode` string for it.

When it's a separate flag, than it's unclear when it's used, when ignored and it becomes hard to understand. When it is its own mode, then there's no problem like this, and it's clear then it's a slightly different mode.

I propose:

* `use-zstd-full`
* `use-zstd-incremental` or `use-zstd-diff`
* define the `use-zstd` and alias to "whatever we think is the best default", and make it `use-zstd-incremental` right now.

@ipetkov

dpc:
> Will try to rebase and debug.

Ah, damn it. `sqlite-sys`'s `build.rs` uses `std::fs::copy` to generate the file by using checked-in copy.

Unfotunately on MacOS that uses `fcopyfile`:

> The `fcopyfile` function preserves the modification, access, and creation timestamps from the source file to the destination file by default.

So the generated file copied the mtime from the existing one, and gets skipped. Bummer. Didn't take long to hit and overlooked issue.


Another reason why `use-zstd-full` will be required sometimes, but also I have a simple solution - if the given deriation doesn't inherit any artifacts, it should just always do the full snapshot.

dpc:
Re: https://github.com/rust-lang/rust/issues/115982

j-baker:
Yep - that's (among other reasons) why I didn't like the look of anything which relies on FS metadata - too easy to mess it up accidentally, especially when cross platform.

inotify might be another viable approach? subscribe to the dir pre-build, watch for events and stash them for later?

dpc:
I think it's OK as is for now. Thanks to using a full mode in deps only step, the worst case (build.rs scripts of 3rd party tools) are virtually never going to get affected by this. And it's always possible to revert to `use-zstd-full`.

It would be possible make a `use-zstd-diff-precise`. During decompression it would store somewhere next to `./target` a list of all paths with their checksums, then during compression use mtime as a first fast check, and if `0`, double-verify the existence, then checksum. Since everything here is basically IO-bound, an extra CPU time would get largely amortized and everything here should parallize well. Hey did I mentioned that I wrote tools like [`rdedup`](https://github.com/dpc/rdedup), [`rblake2sum`](https://github.com/crev-dev/rblake2sum) that suck in data as fast as IO let them? :)

lib/setupHooks/installCargoArtifactsHook.sh
ipetkov: `installCargoArtifactsMode` allows us to only specify one thing, meaning if we want to subtly tweak different installation strategies we'd have to keep tacking on `use-zstd-flag-1`, `use-zstd-flag-2`, `use-zstd-flag-1-and-flag-2` which could get unweildy IMO

ipetkov:
I think for now we can force Darwin builds to do a non-incremental compression until https://github.com/rust-lang/rust/issues/115982 is resolved. We can also follow up with other optimizations later! (I think there's value in landing this in the mean time)

ipetkov:
@dpc I pushed my changes to a separate branch (I had forked off before you had made other changes and I didn't want to necessarily clobber your branch): https://github.com/ipetkov/crane/pull/398

I think I've addressed all the points we've brought up here (for the time being), feel free to take it for a spin

lib/setupHooks/installCargoArtifactsHook.sh
dpc: We're not going to support that many variations, IMO. But if it's neccessary it could be

```
installCargoArtifactsMode = { mode = "use-zstd"; diff = true; };
```

which crane can easily detect and handle, I think.

@ipetkov

ipetkov:
I'm going to close this since https://github.com/ipetkov/crane/pull/398 has landed (though feel free to open a new PR if there's anything else worth changing). Thanks again for the feedback and initial direction here!

PR #392, opened 2023-09-18
flake: add a description

grahamc:
Crane is one of the most popular libraries on FlakeHub, but it won't appear as "featured" on the homepage because there is no description. I thought I'd help fix that if you want it!

## Checklist

- [ ] updated `CHANGELOG.md`


ipetkov:
Heh, thanks for fixing that!

PR #398, opened 2023-09-23
feature: incremental zstd mode (reworked)

ipetkov:
## Motivation
Slightly reworked version of https://github.com/ipetkov/crane/pull/387

## Checklist

- [ ] updated `CHANGELOG.md`


ipetkov:
Note: docs still need to be updated

dpc:
Feel free to squash, etc, I don't care about any attribution, only about the end functionality.

I did a cursory look and i think it works, and I think it has all the fixes from https://github.com/ipetkov/crane/pull/387. I will try it out later.

The only disagreement I have is that an additional flag to force full dump is a meh UX, easy to overlook. If you don't like another `mode` for it, I would suggest `cargoArtifactsInstallMode = { mode = "use-zstd"; full = true; };` . The problem with an extra flag is just that it goes into global scope of options, while it's functionally affecting only this one other argument. I really feel it should just be a paramter to `cargoArtifactsInstallMode`.

Anyway - just an opinion. Will be happy to have it one way or another.

dpc:
@ipetkov Something seems wrong with the way the diff is captured or restored, as the next step attempts to recompile the dependencies.

dpc:
```
fedimint-test-all> decompressing cargo artifacts from /nix/store/f03c5fk7cvza9873w9v51xm3v50kg9b7-fedimint-workspace-build-0.0.1/target.tar.zst to target
fedimint-test-all> configuring
fedimint-test-all> will append /build/source/.cargo-home/config.toml with contents of /nix/store/lj8rblj4isjk3ly2jv27k319xhs6zks2-vendor-cargo-deps/config.toml
```

Seems like it did not restore the the second layer (the diff)?

lib/setupHooks/inheritCargoArtifactsHook.sh
```diff
@@ -17,6 +17,12 @@ inheritCargoArtifacts() {
 
   mkdir -p "${cargoTargetDir}"
   if [ -f "${preparedArtifacts}" ]; then
+
+    local prevArtifactsCandidate="${preparedArtifacts}.prev"
+    if [ -f "${prevArtifactsCandidate}" ]; then
+      inheritCargoArtifacts "${prevArtifactsCandidate}" "$cargoTargetDir"
```
dpc, line 23: This needs to be recursive BTW.

dpc: Oh. I see. It was meant to be

```nix
      inheritCargoArtifacts "${prevArtifactsCandidate}" 
      inheritCargoArtifacts "$cargoTargetDir"
```

lib/setupHooks/installCargoArtifactsHook.sh
```diff
@@ -1,20 +1,46 @@
...
+        >&2 echo "incremental zstd compression not currently supported on Darwin: https://github.com/rust-lang/rust/issues/115982"
+        tar "$@" "${cargoTargetDir}"
+      elif [ -z "${prevArtifacts}" ]; then
+        >&2 echo "no previous artifacts found, compressing and installing full archive"
+        tar "$@" "${cargoTargetDir}"
+      else
+        >&2 echo "linking previous artifacts ${prevArtifacts}"
+        ln -s "${prevArtifacts}" "${dest}.prev"
+        touch -d @${SOURCE_DATE_EPOCH} "${TMPDIR}/.crane.source-date-epoch"
+        tar \
```
dpc, line 28: This shold print something about compressing the diff.

lib/setupHooks/installCargoArtifactsHook.sh
```diff
@@ -1,20 +1,46 @@
...
+        tar "$@" "${cargoTargetDir}"
+      elif [ "$(uname -s)" == "Darwin" ]; then
+        # https://github.com/rust-lang/rust/issues/115982
+        >&2 echo "incremental zstd compression not currently supported on Darwin: https://github.com/rust-lang/rust/issues/115982"
+        tar "$@" "${cargoTargetDir}"
+      elif [ -z "${prevArtifacts}" ]; then
+        >&2 echo "no previous artifacts found, compressing and installing full archive"
+        tar "$@" "${cargoTargetDir}"
+      else
+        >&2 echo "linking previous artifacts ${prevArtifacts}"
```
dpc, line 25: The output here is confusing and might get people to misread the times how long thing took.

`echo "compressing ${cargoTargetDir} to ${dest}"` should happen right before the actual `tar | zstd`, IMO.

lib/setupHooks/inheritCargoArtifactsHook.sh
dpc, line 23: In my change I did `readlink` here, as there are some checks in `inheritCargoArtifacts` that I'm not sure are going to be happy with a symlink. Even if it is working now, it will be easy to introduce (possibly silent, as it will only trigger rebuilds) bug in the future.

checks/default.nix
```diff
@@ -107,6 +107,23 @@ in
     };
   });
 
+  chainedMultiple = pkgs.linkFarmFromDrvs "chainedMultiple" (map
```
dpc, line 110: All these checks are not going to catch any issues. The only reason CI in Fedimint catches the problem is because we have a `cargo/rustc` wrapper that forbids rebuilds at certain point of the chain. Otherwise bugs w.r.t. target inheritance etc. just cause slow rebuilds which is easy to overlook.

lib/setupHooks/installCargoArtifactsHook.sh
```diff
@@ -1,20 +1,46 @@
...
+        ln -s "${prevArtifacts}" "${dest}.prev"
```
dpc, line 26: This (currently) links to the directory containing `target.tar.zst`, not the file itself:

```
fedimint-workspace-build> linking previous artifacts /nix/store/bxlki6y5zn5lgj2dlcxq2jh687qxzxzw-fedimint-deps-0.0.1
```

which will make the `if [ -f $candidate ]; ...` ignore it.

dpc:
@ipetkov Found a bug, I think.

checks/default.nix
dpc: You might just want to copy&paste this wrapper into your testing suite ... . It comes down to https://github.com/fedimint/fedimint/blob/97b3b701aed2dffc556edd12c2159afcf9aa526d/flake.toolchain.nix#L130 and https://github.com/fedimint/fedimint/tree/master/scripts/dev/rustc-wrapper .

It seems that testing that rustc doesn't rebuild stuff is quite important and common.

ipetkov: Ah thanks for the suggestion! We've got a couple "do-not-rebuild" tests but they're pretty manual and annoying to deal with.

Will take a look at incorporating this

dpc:
No stress but this PR is probably like 2 lines change away from being good to go. ;D

ipetkov:
Been busy with travel and other stuff, I'll try to find the time to round it out!

lib/setupHooks/installCargoArtifactsHook.sh
ipetkov: I can tweak the logs here

lib/setupHooks/inheritCargoArtifactsHook.sh
ipetkov: Could you elaborate on what specifically wouldn't be happy with a symlink?

`inheritCargoArtifacts` already does some introspection on whether the input is a directory or not (which should work fine with or without symlinks) so I can't think of a situation where things might fail

dpc: Nothing specific. Just a normal path passes a normal path there, not a symlink. I'm wary of symlinks vs non-symlinks issue. Some tools sometimes do some different things because of them.

But it will probably be OK. Nothing specific.

lib/setupHooks/installCargoArtifactsHook.sh
ipetkov: We were missing the logic for handling other kinds of `.prev` artifacts (including other directories). I added some new logic in if you want to try it out (still need to incorporate the no-rebuild logic in the tests to validate this won't regress)

ipetkov: Simplified this drastically, there's no need to be so hyper generic (it makes reasoning about the inputs really difficult). We can always revisit if its necessary to be more generic

Based on the logs this should be working now!

lib/setupHooks/inheritCargoArtifactsHook.sh
ipetkov: I think this is fine for now (initial testing shows that the symlink is working out fine) so I'll mark this as resolved but happy to revisit if we ever find problems with it

checks/default.nix
ipetkov: @dpc have some questions about how your wrapper script works: i see there's an environment variable which can be flipped to deny any new compilation, makes sense, but how to you check that there are no rebuilds between the `deps-only` and the actual package build?

Example:
1. build deps only with the wrapper, env var not set things build from scratch for the first time
2. build the real workspace source with artifacts from 1, env var not set since we still need to build the workspace
3. Set the env var, rebuild the workspace package with artifacts from 2, nothing should build

This example makes sense to me, but I'm curious if you do (or if you could) check that there aren't any extra rebuilds between steps 1 and 2

dpc: > but how to you check that there are no rebuilds between the `deps-only` and the actual package build?

I don't use it like that. I only use it after I have built whole workspace and I run derivations for tests.

ipetkov: Updated, thanks again for the suggestion!

dpc:
Looks like our CI passes just fine

```
fedimint-test-all> decompressing cargo artifacts from /nix/store/2qw3fjizlfg8zm3c5gln95wqkdr4rvy7-fedimint-deps-0.0.1/target.tar.zst to target
fedimint-test-all> decompressing cargo artifacts from /nix/store/lm76rry8m56k5vx9x3v4mdilg1sb9wvb-fedimint-workspace-build-0.0.1/target.tar.zst to target
fedimint-test-all> configuring
```

All the sizes etc. look OK, no rebuilds.

ipetkov:
Sweet, let's get this merged then!

PR #402, opened 2023-09-29
fix: isolate scope of craneUtils

dpc:
## Motivation

As things are currently it's not feasible to use `craneLib.overrideScope` to override anything that is used by `craneUtils`, because it will affect it as well. Eg. You want to inject some args to every `mkCargoDerivation` be overriding it? Well, now `craneUtils` will try to compile it.

IMO, `craneUtils` is an input to the scope. But I understand that it makes sense to be build with with `craneLib` itself.

There's probably some better way to do it, but I mainly wanted to explain the issue by showing the solution.

## Checklist

- [ ] updated `CHANGELOG.md`

### DIFF
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -1,8 +1,10 @@
 { lib
 , newScope
-  # , craneUtils ? (import ./. { inherit lib newScope; craneUtils = throw "craneUtils mut not depend on itself"; }).craneUtils
-  # , craneUtils ? callPackage ../pkgs/crane-utils { };
-, craneUtilsIsolate ? true
+  # craneUtils to use
+  # when missing, will be generated in a new scope
+  # when null will be generated in the current scope
+  # when set, will be used
+, craneUtils ? (import ./. { inherit lib newScope; craneUtils = null; }).craneUtils
 }:
 
 lib.makeScope newScope (self:
@@ -10,10 +12,11 @@ let
   inherit (self) callPackage;
 
   craneUtils' =
-    if craneUtilsIsolate then
-      (import ./. { inherit lib newScope; craneUtilsIsolate = false; }).craneUtils
+    if craneUtils == null then
+      callPackage ../pkgs/crane-utils { }
     else
-      callPackage ../pkgs/crane-utils { };
+      craneUtils;
+
   internalCrateNameFromCargoToml = callPackage ./internalCrateNameFromCargoToml.nix { };
 in
 {
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -1,11 +1,19 @@
 { lib
 , newScope
+  # , craneUtils ? (import ./. { inherit lib newScope; craneUtils = throw "craneUtils mut not depend on itself"; }).craneUtils
+  # , craneUtils ? callPackage ../pkgs/crane-utils { };
+, craneUtilsIsolate ? true
 }:
 
 lib.makeScope newScope (self:
 let
   inherit (self) callPackage;
 
+  craneUtils' =
+    if craneUtilsIsolate then
+      (import ./. { inherit lib newScope; craneUtilsIsolate = false; }).craneUtils
+    else
+      callPackage ../pkgs/crane-utils { };
   internalCrateNameFromCargoToml = callPackage ./internalCrateNameFromCargoToml.nix { };
 in
 {
@@ -30,7 +38,9 @@ in
   cleanCargoToml = callPackage ./cleanCargoToml.nix { };
   configureCargoCommonVarsHook = callPackage ./setupHooks/configureCargoCommonVars.nix { };
   configureCargoVendoredDepsHook = callPackage ./setupHooks/configureCargoVendoredDeps.nix { };
-  craneUtils = callPackage ../pkgs/crane-utils { };
+
+  craneUtils = craneUtils';
+
   devShell = callPackage ./devShell.nix { };
 
   crateNameFromCargoToml = callPackage ./crateNameFromCargoToml.nix {


dpc:
I pushed (possibly) nicer approach, thought I haven't test it yet. Possibly should get squashed if it was to be merged.

ipetkov:
Could you elaborate on why this change is necessary and what situation it would resolve?

My understanding is that `.overrideScope'` will effectively apply the changes "recursively" (similarly to how a nixpkgs overlay will replace that definition _everywhere_. Namely setting `.overrideScope' (final: prev: { mkCargoDerivation = ...; })` should get applied everywhere, including the definition for `craneUtils`

dpc:
> My understanding is that `.overrideScope'` will effectively apply the changes "recursively" (similarly to how a nixpkgs overlay will replace that definition _everywhere_. Namely setting `.overrideScope' (final: prev: { mkCargoDerivation = ...; })` should get applied everywhere, including the definition for `craneUtils`

That's the problem. E.g. I don't want to change the profile in which `craneUtils`, just because I overwrote `mkCargoDerivation` to add a default `CARGO_PROFILE`. I wanted to do it for the stuff I am compiling, and currently there's no way to separate these. Please see [the discussion of the example](https://github.com/ipetkov/crane/discussions/401) or [the tutorial for the project I'm building on top of crane](https://github.com/rustshop/flakebox/blob/master/docs/building-new-project.md)

ipetkov:
> I wanted to do it for the stuff I am compiling, and currently there's no way to separate these.

@dpc what about

```nix
craneLib.overrideScope' (final: prev: {
  inherit (prev) craneUtils;
  mkCargoDerivation = ...;
})
```

That will effectively apply a new `mkCargoDerivation` everywhere, except use the previous version of `craneUtils` so that the injected changes won't affect it

dpc:
> That will effectively apply a new `mkCargoDerivation` everywhere, except use the previous version of `craneUtils` so that the injected changes won't affect it

Thanks. I think it might work , unless `prev.craneUtils` sneaks in a call to `self.xyz` or something. 

I still think that architecturally `craneUtils` is an input and implementation detail and not the part of the scope returned to the user. It makes sense for it to us it's own `craneLib` to build itself, but it doesn't feel right to force the caller to worry about not breaking it's own building.

dpc:
Actually does `inherit (prev) craneUtils;` even work?

The `prev.craneUtils` is still calling things via `self.xyz` which is provided at the end during final calls, no?

ipetkov:
Hmm I had thought `overrideScope'` worked kind of like overlays but it doesn't quite work like that :thinking: 

What about something more manual like:

```nix
craneLib.overrideScope' (final: prev: {
  mkCargoDerivation = args: (prev.mkCargoDerivation.override { inherit (prev) craneUtils; }) (args // { ... });
})
```

dpc:
> What about something more manual like:
> 
> ```nix
> craneLib.overrideScope' (final: prev: {
>   mkCargoDerivation = args: (prev.mkCargoDerivation.override { inherit (prev) craneUtils; }) (args // { ... });
> })
> ```

I haven't tried (I might at some point), but I don't think it changes anything. At the core scopes are just "`final` is a lazy function that will be provided after everything was overwritten, and you try to use it". So no matter how we twist it `prev.craneUtils` will call the newly overwritten `mkCargoDerivation` to build itself.

ipetkov:
Hmm perhaps `overrideScope'` isn't the right approach for this. Maybe a more straightforward approach would work about better by explicitly replacing things with `.override`. E.g.:

```nix
let
  mkCargoDerivation = { ... }: ...; # custom behavior
  opinionatedCrane = rec {
    buildDepsOnly = craneLib.buildDepsOnly.override {
      inherit mkCargoDerivation;
    };

   buildPackage = craneLib.buildPackage.override {
     inherit buildDepsOnly mkCargoDerivation;
   };

   # etc...
  };
in
craneLib // opinionatedCrane
```

dpc:
Oh, you really thrown a curve-ball at my brain. :D . Do elements of craneLib get `.override` on each of them? Is it because of `callPackage` or where does it happen? :D


Will the user of such enhanced craneLib loose the ability to customize things if they need to, like in vanilla craneLib?

ipetkov:
> Do elements of craneLib get .override on each of them?

Yeah they should I do believe this comes from `callPackage` so try it out!

> Will the user of such enhanced craneLib loose the ability to customize things if they need to, like in vanilla craneLib?

Maybe? I'm not sure (and it will depend on what functionality is desired). You can still `.override` any of the new functions at will, but using something like `.overrideScope'` probably won't work (but as we've seen from testing that's already kinda dodgy for really fine-grained overrides).

There might be another approach which is to use [a fresh `lib.makeScope`](https://github.com/ipetkov/crane/blob/f2143cd27f8bd09ee4f0121336c65015a2a0a19c/lib/default.nix#L5) but pull base/unchanged stuff from `craneLib` and see where that goes

dpc:
> There might be another approach which is to use [a fresh `lib.makeScope`](https://github.com/ipetkov/crane/blob/f2143cd27f8bd09ee4f0121336c65015a2a0a19c/lib/default.nix#L5) but pull base/unchanged stuff from `craneLib` and see where that goes

Intuition tells me this one is a cleaner and more powerful approach. Maybe I could even automatically import all keys and pretend I didn't. I'll look into it. Thanks for the suggestion.

ipetkov:
Worth noting that `lib.makeScope` does make it so `craneLib.newScope` is available, meaning that if you did 

```nix
lib.makeScope craneLib.newScope (self: {
  # ...
});```
 
you'd automatically inherit any definitions in `craneLib` (i.e. this is the same way we are able to refer to stuff in `pkgs`). Haven't tested what other quirks there may be, but hopefully this gives you some ideas to explore!

dpc:
> Worth noting that `lib.makeScope` does make it so `craneLib.newScope` is available, meaning that if you did

It looks like this breaks the interception of `mkCargoDerivation` for me.

Maybe there's some good way to solve it, but at this point I'm piling up complexity, where my goal was to inject 10-20 lines of code in craneLib, not invent my whole new libraries. The more I diverge from crane , the more sense it makes to just fork, add 10 lines of code as a commit and keep rebasing when crane has releases. :shrug:

PR #405, opened 2023-09-29
fix #404 - lib/cargoTarpaulin: lowercase --out xml argument

### LINKED ISSUE #404: cargoTarpaulin: invalid value 'Xml'
It seems like tarpaulin broke their cli parsing by only allowing lowercase `--out` arguments.
```
command cargo tarpaulin --release --skip-clean --out Xml --output-dir /nix/store/...-tarpaulin-0.1.0
error: invalid value 'Xml' for '--out [<FMT>...]'
  [possible values: json, stdout, xml, html, lcov]

  tip: a similar value exists: 'xml'
```

marijanp:
## Motivation
fixes #404 
verified it in one of my projects using the `cargoTarpaulin` derivation

## Checklist

- [ ] updated `CHANGELOG.md`

### DIFF
--- a/lib/cargoTarpaulin.nix
+++ b/lib/cargoTarpaulin.nix
@@ -1,9 +1,12 @@
 { mkCargoDerivation
 , cargo-tarpaulin
+, lib
 }:
-
+let
+  xml = if lib.versionOlder cargo-tarpaulin.version "0.27.0" then "Xml" else "xml";
+in
 { cargoExtraArgs ? ""
-, cargoTarpaulinExtraArgs ? "--skip-clean --out Xml --output-dir $out"
+, cargoTarpaulinExtraArgs ? "--skip-clean --out ${xml} --output-dir $out"
 , ...
 }@origArgs:
 let
--- a/docs/API.md
+++ b/docs/API.md
@@ -625,7 +625,7 @@ Except where noted below, all derivation attributes are delegated to
   - Default value: `""`
 * `cargoTarpaulinExtraArgs`: additional flags to be passed in the cargo
   tarpaulin invocation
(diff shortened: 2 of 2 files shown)


lib/cargoTarpaulin.nix
```diff
@@ -3,7 +3,7 @@
 }:
 
 { cargoExtraArgs ? ""
-, cargoTarpaulinExtraArgs ? "--skip-clean --out Xml --output-dir $out"
+, cargoTarpaulinExtraArgs ? "--skip-clean --out xml --output-dir $out"
```
ipetkov, line 6: The 23.05 branch currently has cargo-tarpaulin `0.25.2` which accepts the `Xml` format, while the version in nixpkgs-unstable is at `0.27.0` so something must have changed there...

We could introspect the cargo-tarpaulin version and set the default here accordingly (at least until the 23.11 is released)

(Also docs are great as is, we don't need to highlight this handling there)

```nix
let
  xml = if lib.versionOlder cargo-tarpaulin.version "0.27.0" then "Xml" else "xml";
in
{ cargoExtraArgs ? ""
, cargoTarpaulinExtraArgs ? "--skip-clean --out ${xml} --output-dir $out"
, ...
}:
```

ipetkov (approved):
Thanks for fixing this! Have one comment about allowing things to work on the latest stable release branch, but happy to merge once addressed!

lib/cargoTarpaulin.nix
marijanp: Thanks for taking a look, I've added the version check.

ipetkov (approved):

PR #406, opened 2023-09-29
Fix postInstall name

pingiun:
## Motivation
The link to the manual specifies that it should be postInstall, not postInstallPhase

## Checklist

- [x] Simple docs update


ipetkov:
Thanks for fixing this!

Issue #394, opened 2023-09-22
How to bypass cleanCargoToml for specific dependencies?


danjl1100:
I am getting a rustc error when building a crate with any dependencies that themselves depend on [document-features](https://lib.rs/crates/document-features)  (e.g. many parts of [embassy](https://github.com/embassy-rs/embassy)).

The issue seems to be that the `document_features!` proc-macro depends on "doc comments" being present in Cargo.toml at build time, whereas crane is stripping out comments (for better cache hits?).

I experimented with overriding crane's `cleanCargoToml` attribute, but it did not seem to apply to dependencies of the crate being built.   See further notes at the [reproducible example](https://github.com/danjl1100/example_uses_dep_using_document_features), which uses a [minimal dependency](https://github.com/danjl1100/example_uses_document_features) that itself depends on `document-features`.

<details>
<summary>Summary of rustc error message</summary>

More details at the README for the [reproducible example](https://github.com/danjl1100/example_uses_dep_using_document_features).

```
$ cargo c
+ command cargo c
    Finished dev [unoptimized + debuginfo] target(s) in 0.03s

$ nix build
warning: Git tree '/home/user/the_bin' is dirty
error: builder for '/nix/store/bvizj8r1dbz8941pbvqr1s0s0qpr2w6z-the_bin-deps-0.1.0.drv' failed with exit code 101;
       last 10 log lines:
       >     Checking example_uses_document_features v0.1.0 (https://github.com/danjl1100/example_uses_document_features.git#a1ba56e5)
       > error: Could not find documented features in Cargo.toml
       >  --> /nix/store/680bw6m3wbvjim5qsymmllcbk61pf7v5-vendor-cargo-deps/6324285ea70798e13c0a2ff48894e089c59e1e9b1901e701033e48ef46db6c00/example_uses_document_features-0.1.0/src/lib.rs:2:10
       >   |
       > 2 | #![doc = document_features::document_features!()]
       >   |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
       >   |
       >   = note: this error originates in the macro `document_features::document_features` (in Nightly builds, run with -Z macro-backtrace for more info)
       >
       > error: could not compile `example_uses_document_features` (lib) due to previous error
       For full logs, run 'nix log /nix/store/bvizj8r1dbz8941pbvqr1s0s0qpr2w6z-the_bin-deps-0.1.0.drv'.
[...]

PR #407, opened 2023-09-30
Replace `toml` dependency with `toml_edit` to preserve comments

danjl1100:
## Motivation

As discussed in #394, crane currently uses the [`toml` crate] dependency in `crane-resolve-workspace-inheritance.rs` in order to merge the workspace root Cargo.toml with the current crate's Cargo.toml. This script is run on all git dependencies of the built crate.

One unfortunate side effect of using the [`toml` crate] is that all comments are removed from Cargo.toml.  This is an issue if proc macros or build scripts are expecting to read non-TOML metadata from Cargo.toml (e.g. [`document-features`] used by some embedded libraries, such as [`embassy-stmf32`](https://github.com/embassy-rs/embassy/tree/main/embassy-stm32))

### Solution

This PR adapts the existing `crane-resolve-workspace-inheritance.rs` script to use [`toml_edit`] instead of the [`toml` crate], to preserve comments.

#### Remarks
- There are two kinds of Tables in toml_edit, so some complexity (e.g. `trait TableLikeExt`) is needed to merge any combination of tables.
- The TOML key/value formatting cleanup is not strictly required, but I included this to keep the current smoke test mostly unchanged.   One possible alternative is to use the [`toml` crate] as a dev-dependency, to only check the TOML is equal and avoid string equality; then the `fmt` module could be removed.

I tested this on my minimal example from #394, and it appears to work correctly.  (reference https://github.com/danjl1100/example_uses_dep_using_document_features/commit/7cc5174ed13868d49eef1900e88e3c7157b112e4)

## Checklist

- [x] updated `CHANGELOG.md`

[`toml` crate]: https://lib.rs/crates/toml
[`toml_edit`]: https://lib.rs/crates/toml_edit
[`document-features`]: https://lib.rs/crates/document-features
[`embassy-stmf32`]: https://github.com/embassy-rs/embassy/tree/main/embassy-stm32

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   ([#376](https://github.com/ipetkov/crane/pull/376))
 * Replaced various internal usages of `runCommandLocal` with `runCommand` for
   more optimal behavior when downloading cached artifacts
+* Fixed a bug when a git dependency crate relies on reading non-TOML metadata
+  from Cargo.toml at build time.
+  ([#407](https://github.com/ipetkov/crane/pull/407))
 
 ## [0.13.1] - 2023-08-22
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
-### Changed
-* Fixed a bug when a git dependency crate relies on reading non-TOML metadata
-  from Cargo.toml at build time.
-  ([#407](https://github.com/ipetkov/crane/pull/407))
+### Fixed
+* Fixed handling of Cargo workspace inheritance for git-dependencies where said
+  crate relies on reading non-TOML metadata (i.e. comments) from its Cargo.toml
+  at build time. ([#407](https://github.com/ipetkov/crane/pull/407))
 
 ## [0.14.1] - 2023-09-23
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Changed
+* Fixed a bug when a git dependency crate relies on reading non-TOML metadata
+  from Cargo.toml at build time.
+  ([#407](https://github.com/ipetkov/crane/pull/407))
+
 ## [0.14.0] - 2023-09-21
 
 ### Added
@@ -31,9 +36,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   ([#376](https://github.com/ipetkov/crane/pull/376))
 * Replaced various internal usages of `runCommandLocal` with `runCommand` for
   more optimal behavior when downloading cached artifacts
-* Fixed a bug when a git dependency crate relies on reading non-TOML metadata
-  from Cargo.toml at build time.
-  ([#407](https://github.com/ipetkov/crane/pull/407))
 
 ## [0.13.1] - 2023-08-22
 
--- a/pkgs/crane-utils/src/bin/crane-resolve-workspace-inheritance.rs
+++ b/pkgs/crane-utils/src/bin/crane-resolve-workspace-inheritance.rs
@@ -133,39 +133,18 @@ fn try_as_table_like_mut(item: &mut Item) -> Option<&mut dyn toml_edit::TableLik
 
 /// Merge the specified `cargo_toml` and workspace `root` if both are tables
 fn try_merge_cargo_tables(cargo_toml: &mut Item, root: &Item) {
-    match (cargo_toml, root) {
-        (
-            Item::Table(p), //
-            Item::Table(wp),
-        ) => {
-            merge_cargo_tables(p, wp);
-        }
-        (
-            Item::Value(toml_edit::Value::InlineTable(p)), //
-            Item::Table(wp),
-        ) => {
-            merge_cargo_tables(p, wp);
-        }
-        (
-            Item::Table(p), //
-            Item::Value(toml_edit::Value::InlineTable(wp)),
-        ) => {
-            merge_cargo_tables(p, wp);
-        }
-        (
-            Item::Value(toml_edit::Value::InlineTable(p)),
-            Item::Value(toml_edit::Value::InlineTable(wp)),
-        ) => {
-            merge_cargo_tables(p, wp);
-        }
-        _ => {}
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -31,6 +31,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   ([#376](https://github.com/ipetkov/crane/pull/376))
 * Replaced various internal usages of `runCommandLocal` with `runCommand` for
   more optimal behavior when downloading cached artifacts
+* Fixed a bug when a git dependency crate relies on reading non-TOML metadata
+  from Cargo.toml at build time.
+  ([#407](https://github.com/ipetkov/crane/pull/407))
```
ipetkov, line 36: This should go in a fresh `### Changed` section after `## Unreleased` (since the 0.14.0 tag has already been published) :smile:

pkgs/crane-utils/src/bin/crane-resolve-workspace-inheritance.rs
```diff
@@ -265,6 +471,7 @@ mod tests {
             version = "some version"
 
             [workspace.dependencies]
+            # workspace comments are not copied
```
ipetkov, line 474: FWIW I'm totally fine to keep all comments in these `Cargo.toml`s (that is to say we don't need to go out of our way to strip them out, i.e. drop the `fmt` module entirely).

Stripping comments is desired when handling building/developing a project itself (to avoid accidental rebuilds), but keeping them in for any git-deps doesn't hurt since the only time they would change is if the dependency's commit is updated in the lock file (which would end up rebuilding anyway)!

pkgs/crane-utils/src/bin/crane-resolve-workspace-inheritance.rs
```diff
@@ -105,90 +96,300 @@ fn merge(cargo_toml: &mut toml::Value, root: &toml::Value) {
...
+        Item::Value(toml_edit::Value::InlineTable(w)) => Some(w),
+        _ => None,
+    }
+}
+
+/// Merge the specified `cargo_toml` and workspace `root` if both are tables
+fn try_merge_cargo_tables(cargo_toml: &mut Item, root: &Item) {
+    match (cargo_toml, root) {
+        (
+            Item::Table(p), //
```
ipetkov, line 138: Missing comment? Or is this to nudge rustfmt?

pkgs/crane-utils/src/bin/crane-resolve-workspace-inheritance.rs
```diff
@@ -105,90 +96,300 @@ fn merge(cargo_toml: &mut toml::Value, root: &toml::Value) {
...
+    match item {
+        Item::Table(w) => Some(w),
+        Item::Value(toml_edit::Value::InlineTable(w)) => Some(w),
+        _ => None,
+    }
+}
+
+/// Merge the specified `cargo_toml` and workspace `root` if both are tables
+fn try_merge_cargo_tables(cargo_toml: &mut Item, root: &Item) {
+    match (cargo_toml, root) {
```
ipetkov, line 136: Would it be simpler to use `try_as_table_like{,mut}` instead of matching on every combination of the variants? Or is the goal to avoid the dynamic dispatch?

pkgs/crane-utils/src/bin/crane-resolve-workspace-inheritance.rs
```diff
@@ -105,90 +96,300 @@ fn merge(cargo_toml: &mut toml::Value, root: &toml::Value) {
...
+            merge_tables(dest, additional);
+        }
+        Item::None => {}
+        Item::ArrayOfTables(additional) => {
+            if let Item::ArrayOfTables(dest) = dest {
                 dest.extend(additional);
             } else {
-                // Override dest completely if types don't match
-                *dest = toml::Value::Array(additional);
+                *dest = Item::ArrayOfTables(additional);
```
ipetkov, line 229:
```suggestion
                 // Override dest completely if types don't match
                *dest = Item::ArrayOfTables(additional);
```

Just for consistency/clarity

ipetkov (commented):
Thanks for taking this on! Direction looks good, just have a few minor questions

CHANGELOG.md
danjl1100: Ah, I see what you mean!  Updated the changelog to fix.

pkgs/crane-utils/src/bin/crane-resolve-workspace-inheritance.rs
danjl1100: I updated the comment to clarify: the way we are performing the merge, the "surrounding decoration" is only copied over values that need to be added.   Since this top-level comment is on a key "foo" which is in common with cargo_toml, the keys are merged and the comment is taken from the "cargo_toml" side.

I also removed the `fmt` module in a separate commit, to show how the resulting `smoke_test` truth has to change.  Do you think this is OK as-is (string comparison), or should I try using the `toml` crate as a dev-dependency to compare just TOML content, to restore more sane spacing?

pkgs/crane-utils/src/bin/crane-resolve-workspace-inheritance.rs
danjl1100: I updated the comment to clarify: the way we are performing the merge, the "surrounding decoration" is only copied over values that need to be added.   Since this top-level comment is on a key "foo" which is in common with cargo_toml, the keys are merged and the comment is taken from the "cargo_toml" side.

I also removed the `fmt` module in a separate commit, to show how the resulting `smoke_test` truth has to change.  Do you think this is OK as-is (string comparison), or should I try using the `toml` crate as a dev-dependency to compare just TOML content, to restore more sane spacing?

pkgs/crane-utils/src/bin/crane-resolve-workspace-inheritance.rs
danjl1100: This was a nudge to rustfmt, but removed in simplifying to use dyn instead, per the other comment.

pkgs/crane-utils/src/bin/crane-resolve-workspace-inheritance.rs
danjl1100: Updated, thanks! :+1:

pkgs/crane-utils/src/bin/crane-resolve-workspace-inheritance.rs
ipetkov: Ah thanks for clarifying! Yes I think it's totally fine to keep as is for now (even with the non-standard spacing, thanks for noting that in a comment). We can always go with the `toml`-dev-dependency approach later if this code needs to change!

ipetkov (approved):

PR #410, opened 2023-10-02
fix: `cargo docs` breaking in `buildDepsOnly` of workspaces

### LINKED ISSUE #312: Duplicate `crane-dummy` after upgrading crane to v0.12.1
```
fedimint-workspace-deps> error: document output filename collision                                                                    
fedimint-workspace-deps> The bin `crane-dummy` in package `fedi-fedimintd v0.1.0 (/build/source/fedimintd)` has the same name as the bin `crane-dummy` in package `fedi-fedimint-cli v0.1.0 (/build/source/fedimint-cli)`.         
```

dpc:
## Motivation

Fix #312 by using a different name for dummy binary in every package (derived from the package name itself). This should make `cargo doc` happy. I tested lightly and it seems to work.

## Checklist

- [ ] updated `CHANGELOG.md`


ipetkov:
Thanks for the PR! Is it possible to add a test for this to avoid future regressions?

dpc:
Now that I have some idea how the tests work, I can try adding it later today.

BTw. I don't understand the test failure. Something about https://github.com/ipetkov/crane/blob/581245bf1233d6f621ce3b6cb99224a948c3a37f/checks/default.nix#L432  maybe? I guess I'll look at it later as well.

ipetkov:
We have some tests that `cleanCargoToml` outputs things as expected. Since this changes the name used for dummy binaries those tests might need to be updated as well (I'd recommend building the failing test locally to help with debugging)

dpc:
Fixed, added test, made sure test failed before, rebased, pushed. @ipetkov

ipetkov (approved):
Thanks for adding the test, LGTM!

Issue #417, opened 2023-10-11
Binary compiled with `dontStrip = true` crashes


simonzkl:
I was trying to add debug symbols to my `release` binary and I noticed the binary compiled with crane and `dontStrip = true` crashes on startup.

**Example project:** https://github.com/simonzkl/crane-dont-strip

**Steps to reproduce:**
1. `nix run -L`
2. Process exits with `Killed: 9`

**Using `cargo build` in `nix-shell` works fine:**
1. `nix develop -c -- cargo run --release`
2. Application exits with exit code 0

Note that this only seems to happen when you link with external dependencies like `openssl`.

Setting `doNotRemoveReferencesToVendorDir = true` fixes the issue, so it looks like crane patches something that makes MacOS freak out.

I only tested this on `aarch64-darwin`.

dpc:
Hey. We have hit the same issue. I don't think it's a crane issue though. Probably some weird-o-security thing on Macs.

I vaguely recall that upgrading rustc fixed it too.

dpc:
> Possibly something to do with codesigning on darwin?

Yes. Something like that. I'm not an expert in unfree OSes, but I had to deal with SIPs and other problems around it in the past.


> Rust is `rustc 1.72.0` in the example above.

I might have been wrong.

Seems like we still have

```
dontStrip = !pkgs.stdenv.isDarwin; 
```

simonzkl:
Setting `doNotRemoveReferencesToVendorDir = true` fixes the issue, so it looks like crane patches something it shouldn't, which was previously stripped out by `dontStrip = false`. Will update the issue.

dpc:
Ooohhhhh. I didn't know that. OK then. Crane's fault. I think crane is using binary substitution to override some debug paths that would link the binaries with build dir artifact, making Nix consider vendored source code a Nix dependency for your binaries.

I guess it trips some integrity checks / checksums somewhere?

Oh that might be a PITA to undo. THough maybe crane should just `cp -aL` these files into the build env, at least on MacOS / optionally?

simonzkl:
Ok I can confirm it's just a signing issue. I added this to my `packageBuild`:
```
postFixup = ''
  remove-references-to -t ${pkgs.hello} $out/bin/*
'';
```

This triggers the codesigning hook here: https://github.com/NixOS/nixpkgs/blob/5dcafc36852cd8d96127f5cbfe42918bf262214e/pkgs/build-support/remove-references-to/default.nix#L27-L29

[...]

PR #418, opened 2023-10-11
removeReferencesToVendoredSources: sign aarch64-darwin binaries

simonzkl:
## Motivation

Fixes https://github.com/ipetkov/crane/issues/417

On aarch64-darwin we need to sign binaries. Usually this is done automatically, but it seems the signature is invalidated when we patch binaries using `removeReferencesToVendoredSources`.

## Checklist

- [x] updated `CHANGELOG.md`


ipetkov:
Hi @simonzkl thanks for the PR and the test case!

I think there's a simpler way of solving this problem: removing references later in the build (after signing has had a chance to run)! It looks like nixpkgs will perform signing as a fixup hook. Currently references are removed [as a postInstall step](https://github.com/ipetkov/crane/blob/f2143cd27f8bd09ee4f0121336c65015a2a0a19c/lib/setupHooks/removeReferencesToVendoredSourcesHook.sh#L35) but if we changed that to a post fixup hook it might work out?

simonzkl:
I could be wrong but my understanding is that signing runs as part of the clang build when you run `cargo build`, not as a hook of the `buildPackage` derivation. So I'm not sure if we actually have any control over this.

simonzkl:
@ipetkov actually you're right. Moving `removeReferencesToVendoredSources` to `fixupHooks` does solve the issue. I updated the PR.

ipetkov (approved):
Looking good, thanks again for the fix!

ipetkov:
Uh oh, we've got a regression! Actually it looks like the hook isn't running at all, maybe we need to specify `postFixupHooks`?

simonzkl:
Ah sorry, it seems I fixed the issue by not running the hook at all. Changing it to `preFixupHooks`/`postFixupHooks` reintroduces the issue. I think my initial understanding that the signing part runs as part of `cargo build` was correct so this is not going to work. If you looks at `remove-references-to`, they manually run the signing hook, so I think we need to do the same.

simonzkl:
I reverted back to the old approach, but I'm happy to do it differently if there's a better idea. Should we revert `postFixupHooks` back to `postInstallHooks`? Technically this is closer to a fixup hook, but it is a breaking change I suppose.

ipetkov:
Ah I apologize for sending you down the wrong path, I had mistakenly assumed the signing was always being done by the stdenv and we were somehow clobbering it, when in reality we have to explicitly sign _because_ we patch the binary afterwards?

Anyway this seems like the appropriate path (at least it fixes the issues even if I'm still fuzzy on the details)

> Should we revert `postFixupHooks` back to `postInstallHooks`? Technically this is closer to a fixup hook, but it is a breaking change I suppose.

Yes let's revert this (also yay no longer a breaking change!) we can circle back to the change log when this is ready to merge

lib/setupHooks/removeReferencesToVendoredSourcesHook.sh
```diff
@@ -26,13 +26,23 @@ removeReferencesToVendoredSources() {
...
 }
 
 if [ -n "${doNotRemoveReferencesToVendorDir-}" ]; then
   echo "removeReferencesToVendoredSources disabled"
 elif [ -n "${cargoVendorDir-}" ]; then
-  postInstallHooks+=(removeReferencesToVendoredSources)
+  signingUtils="@signingUtils@"
+  if [ -n "${signingUtils}" ]; then
+    source "${signingUtils}"
+  fi
```
ipetkov, line 43: Couple comments:

1. We should unconditionally source this file and always have the functions available (I peeked at the sources, there's no automatic behavior just function definitions). One way we can avoid the checks completely is use `@sourceSigningUtils@` in here and in the derivation definition we can conditionally replace that string with `"source ${signingUtils}`
2. That source should be done outside of these conditional checks (that way if someone manually calls `removeReferencesToVendoredSources` without `$cargoVendorDir` being set it won't crash)

lib/setupHooks/removeReferencesToVendoredSourcesHook.sh
```diff
@@ -26,13 +26,23 @@ removeReferencesToVendoredSources() {
 
       echo -n '\)!@storeDir@/eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee!g'
     )
+
+    # Sign modified binaries on aarch64-darwin
+    if [ -n "${signingUtils}" ]; then
+      signIfRequired "${installedFile}"
+    fi
```
ipetkov, line 33: 1. We should replace the conditional check with a new `doNotSign` variable (or similarly named) to allow it to be easily disabled if necessary
2. Similar to the other comment, we can tweak the placeholder variable to skip this stanza entirely rather than checking if `signingUtils` are set (if any derivation happens to have that variable set we might get confused here)

docs/API.md
```diff
@@ -1538,4 +1538,4 @@ sources themselves. It takes two positional arguments:
 **Automatic behavior:** if `cargoVendorDir` is set and
 `doNotRemoveReferencesToVendorDir` is not set, then
 `removeReferencesToVendoredSources "$out" "$cargoVendorDir"` will be run as a
-post install hook.
+`fixup` hook.
```
ipetkov, line 1541: Leaving a reminder that we should revert this line

CHANGELOG.md
```diff
@@ -6,6 +6,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Changed
+* **Breaking** (technically): `removeReferencesToVendoredSources` will now run
+  later in the derivation as a `fixup` hook and not as a `postInstall` hook. If
+  using `dontFixup = true;` or if you rely on precisely when it runs, you can
+  explicitly configure `removeReferencesToVendoredSources` to be run where
+  appropriate.
```
ipetkov, line 14: Leaving a reminder to fix the changelog here

lib/setupHooks/removeReferencesToVendoredSourcesHook.sh
szlend: I don't think always sourcing `signingUtils` is a good idea, unless it's replaced with an empty file on non-aarch64-darwin system. It will pull in darwin specific binaries as build dependencies.

szlend: `@sourceSigningUtils@` is a good idea though.

ipetkov:
Gonna restart the workflow once the darwin examples cache (looks like we're stuck compiling gcc atm): https://github.com/ipetkov/crane/actions/runs/6525453459/job/17718153027

PR #426, opened 2023-10-15
ci: only build uncached derivations

ipetkov:
## Motivation
Even when a derivation is cached, `nix build` will still download it. I'm not 100% sure if `nix flake check` still downloads cached derivations or not, but there's no harm in ensuring we never download cached derivations by using `nix-eval-jobs`

## Checklist

- [ ] updated `CHANGELOG.md`


ci/fast-flake-check.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+curSystem=$(nix eval --raw --impure --expr builtins.currentSystem)
+
+nixEvalJobsArgs=()
```
dpc, line 4: **question:** Why does it have to be such a long and involved process to do such a popular and needed thing...

ipetkov: If you find out let me know cuz I'm also curious!

PR #440, opened 2023-10-20
cargoLib.cargoDeny: init

aidalgol:
## Motivation

Add `craneLib.cargoDeny`, analogous to `craneLib.cargoAudit` but for [`cargo deny`](https://github.com/EmbarkStudios/cargo-deny), a tool for auditing the licenses and provenance of your crate's dependency graph.

This tool can be used as a replacement for `cargo audit`, but unlike `cargo audit`, which operates on the `Cargo.lock` file, `cargo deny` operates on the `Cargo.toml` file, and so the `craneLib.cargoAudit` derivation has to be rebuilt on any source change, rather than only when the lock file changes.  Because of this, `craneLib.cargoAudit` does *not* run the security-advisory `audit` check.

## Checklist

- [x] updated `CHANGELOG.md`

~~While I was able to follow the various `craneLib` functions well enough to write `cargoDeny`, I have had more difficulty understanding the tests (under `checks/`), and I will need some assistance writing tests for `craneLib.cargoDeny`.~~


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   true`.
 * All dependencies (outside of `nixpkgs`) have been dropped from the (main)
   flake.lock file so they do not pollute downstream projects' lock files.
+* `cargoDeny` added for running [`cargo-deny`](https://github.com/EmbarkStudios/cargo-deny).
```
ipetkov, line 16: This should go in a fresh `### Added` section above `### Changed`

docs/API.md
```diff
@@ -337,6 +337,52 @@ environment variables during the build, you can bring them back via
...
+file, meaning it will be re-run when any source file changes, rather than only
+when dependencies change.
+
+Except where noted below, all derivation attributes are delegated to
+`mkCargoDerivation`, and can be used to influence its behavior.
+* `buildPhaseCargoCommand` will be set to run
+  `cargo deny --offline check bans licenses sources`
+  in the workspace.
+* `pnameSuffix` will be set to `"-deny"`
+* `src` will be filtered to only keep `Cargo.lock` files
```
ipetkov, line 358:
```suggestion

```

examples/quick-start/flake.nix
```diff
@@ -98,6 +98,11 @@
             inherit src advisory-db;
           };
 
+          # Audit licenses
+          my-crate-lib = craneLib.cargoAudit {
```
ipetkov, line 102:
```suggestion
          my-crate-lib = craneLib.cargoDeny {
```

lib/cargoDeny.nix
```diff
@@ -0,0 +1,26 @@
...
+let
+  args = builtins.removeAttrs origArgs [
+    "cargoDenyExtraArgs"
+    "cargoExtraArgs"
+  ];
+in
+mkCargoDerivation (args // {
+  buildPhaseCargoCommand = ''
+    cargo --offline ${cargoExtraArgs} \
+      deny ${cargoDenyExtraArgs} check bans licenses sources
```
ipetkov, line 20: Maybe we should make `bans licenses sources` the default value of `cargoDenyExtraArgs` so that consumers can easily change it if they want to test for something else

docs/API.md
```diff
@@ -337,6 +337,52 @@ environment variables during the build, you can bring them back via
...
+
+#### Required attributes
+* `cargoArtifacts`: A path (or derivation) which contains an existing cargo
+  `target` directory, which will be reused at the start of the derivation.
+  Useful for caching incremental cargo builds.
+  - This can be prepared via `buildDepsOnly`
+  - Alternatively, any cargo-based derivation which was built with
+    `doInstallCargoArtifacts = true` will work as well
```
ipetkov, line 366: Do we actually need artifacts or does `cargo-deny` only need the source? (If they aren't needed, we would also need to change `cargoDeny` to set `cargoArtifacts = null;`)

ipetkov (commented):
> While I was able to follow the various craneLib functions well enough to write cargoDeny, I have had more difficulty understanding the tests (under checks/), and I will need some assistance writing tests for craneLib.cargoDeny.

Sorry the test suite is a little messy! I think we can get started with just a single test which runs on the `./simple` crate. We don't need to test that `cargo-deny` actually catches any specific issues, just that it's running successfully!

docs/API.md
aidalgol: I first tried with all combinations of `cargoArtifacts = null`, `cargoVendorDir = null`, and `doInstallCargoArtifacts = false`; having any of those set in the attrset passed to `mkCargoDerivation` caused the `cargo deny` command to try fetching the crates package index.

ipetkov: We only need `cargoArtifacts = null`; `cargoVendorDir` is okay to leave blank (it will automatically be set if not defined) since we actually need it available to check the dependency licenses

lib/cargoDeny.nix
```diff
@@ -0,0 +1,27 @@
+{ cargo-deny
+, lib
+, mkCargoDerivation
+}:
+
+{ cargoDenyExtraArgs ? ""
+, cargoDenyChecks ? ["bans" "licenses" "sources"]
```
ipetkov, line 7: We should keep this as a string (and not a list of strings). The caller can format it however they'd like but it simplifies things to not need to wrap into a list if there's just one value to pass in

docs/API.md
aidalgol: So the magic combination is `cargoArtifacts = null` and `doInstallCargoArtifacts = false`.  With just `cargoArtifacts = null`, the install phase kicks in right after the `cargo deny` command completes.

Tail end of the Nix log.
```console
++ command cargo --offline deny check bans licenses sources
bans ok, licenses ok, sources ok
@nix { "action": "setPhase", "phase": "installPhase" }
installing
installing target to /nix/store/r3pimy93x88wcnzbli5c02mp9h85qx2n-simple-deny-0.1.0
mv: cannot stat 'target': No such file or directory
```

ipetkov (approved):
Looks good, thanks! Just pushed a few minor fixes

Issue #446, opened 2023-11-01
Build with --store flag fails due to missing Cargo.lock file


tfkhim:
Hey,

first of all I would like to thank you for the great work with crane.

I stumbled across a bug when I try to build my Rust project with the `--store` flag of the `nix` CLI. I get the following error:

    builder for '/nix/store/y7chrz9v1qzj7iqkfkjn5f72804bv966-sway-workspace-extras-deps-0.1.1.drv' failed with exit code 101;
       last 10 log lines:
       > updateAutotoolsGnuConfigScriptsPhase
       > configuring
       > will append /build/source/.cargo-home/config.toml with contents of /nix/store/dys8y52wsl6q2c3fxc29iaqakrrpdzc9-vendor-cargo-deps/config.toml
       > default configurePhase, nothing to do
       > building
       > ++ command cargo --version
       > cargo 1.72.0
       > ++ command cargo check --release --locked --all-targets
       > error: the lock file /build/source/Cargo.lock needs to be updated but --locked was passed to prevent this

It seems like the `Cargo.lock` file is missing in the [cleaned-mkDummySrc](https://github.com/ipetkov/crane/blob/master/lib/mkDummySrc.nix#L110) store entry. If I run the build without the `--store` flag the `Cargo.lock` file is present. 

Right now I set the `cargoLock` option to work around this issue. But according to the examples in the docs this shouldn't be necessary if the lock file is part of the project. Fixing this issues might be also helpful for others because the `nixos-install` script uses the `--store` flag under the hood. Actually, trying to install my system configuration with `nixos-install`was how I found this problem.

You should be able to reproduce this with my [repo](https://github.com/tfkhim/sway-workspace-extras). If you remove the [`cargoLock`](https://github.com/tfkhim/sway-workspace-extras/blob/main/flake.nix#L39) option the above error should show up when you run

    nix build --store "$PWD/test-store"

I think you can reproduce this with any project using crane. But I didn't test this, yet.

I didn't have time to dig deeper and come up with a PR, yet. I plan to continue with the analysis next week. But maybe someone already has a clue what is going wrong and how to fix the issue.

ipetkov:
Hi @tfkhim thanks for the report, I'm able to reproduce this as well!

[...]

PR #448, opened 2023-11-05
mkDummySrc: fix cleaning to work with Nix store overrides

ipetkov:
## Motivation
* Turns out if a build is invoked with `--store` then the source cleaning filter will observe paths rooted at the alternative store which breaks out previous string handling (and results in incorrectly ignoring files which should be included)

Fixes https://github.com/ipetkov/crane/issues/446

## Checklist

- [x] updated `CHANGELOG.md`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 gcroot
 result*
 target/
+/extra-tests/*/alt-store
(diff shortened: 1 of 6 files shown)


tfkhim:
Thank you for the fast fix :+1:

PR #460, opened 2023-11-26
examples: avoid re-instantiating nixpkgs where possible

ipetkov:
## Motivation
* i.e. if we neither set localSystem/crossSystem, nor add overlays, we can directly utilize `nixpkgs.legacyPackages.${system}` directly and avoid re-evaluating nixpkgs an additional time

## Checklist

- [ ] updated `CHANGELOG.md`

### DIFF
--- a/examples/alt-registry/flake.nix
+++ b/examples/alt-registry/flake.nix
@@ -17,9 +17,7 @@
   outputs = { self, nixpkgs, crane, flake-utils, ... }:
     flake-utils.lib.eachDefaultSystem (system:
       let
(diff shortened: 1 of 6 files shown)


ipetkov:
Yes, my understanding is that there are better opportunities for evaluation caching, especially since evaluating `import nixpkgs { system = "..."; }` can be slow with how large it is.

https://discourse.nixos.org/t/using-nixpkgs-legacypackages-system-vs-import/17462

dpc:
Thanks! That is super-useful to know and I actually work in setups in which this might be a huge help.

Issue #480, opened 2023-12-12
Dependencies using `lints` cannot be built


rvolosatovs:
Rust 1.74 added support for `lints` attribute in Cargo manifest (https://blog.rust-lang.org/2023/11/16/Rust-1.74.0.html#lint-configuration-through-cargo), which causes incompatibility it seems, e.g.:
```
failed to update replaced source https://github.com/bytecodealliance/wasmtime?branch=release-16.0.0#759aa585
       >
       > Caused by:
       >   failed to parse manifest at `/nix/store/5qrfydk8pcql8ivlx568a2s6asa40y7v-vendor-cargo-deps/6fe5d6a99257bab41b19f2932c7a0b6aa9306e7fabbef3b691e1c6cb8654cbad/wasmtime-16.0.0/Cargo.toml`
       >
       > Caused by:
       >   error inheriting `lints` from workspace root manifest's `workspace.lints`
```

rvolosatovs:
It's very likely, however, that the issue is that the package is not compatible with `cargo vendor`, I'll have to take a better look tomorrow

ipetkov:
Hi @rvolosatovs thanks for the report! I think we need to teach [`crane-resolve-workspace-inheritance.rs`](https://github.com/ipetkov/crane/blob/62fc1a0cbe144c1014d956e603d56bf1ffe69c7d/pkgs/crane-utils/src/bin/crane-resolve-workspace-inheritance.rs#L93-L97) to handle merging in the `[lints]` table which it knows nothing about today

PR #481, opened 2023-12-13
feat: support workspace `lints`

rvolosatovs:
## Motivation

Closes https://github.com/ipetkov/crane/issues/480 - there could probably be a nicer way to handle this, but I went with the simplest approach I could come up with - note that it appears that `lints.workspace = true` does not support per-crate overrides of the workspace setting at this time, so we don't actually need to merge anything within the crates themselves.

## Checklist

- [x] updated `CHANGELOG.md`

### DIFF
--- a/pkgs/crane-utils/src/bin/crane-resolve-workspace-inheritance.rs
+++ b/pkgs/crane-utils/src/bin/crane-resolve-workspace-inheritance.rs
@@ -90,20 +90,29 @@ fn merge(cargo_toml: &mut toml_edit::Document, root: &toml_edit::Document) {
         };
 
     // https://doc.rust-lang.org/cargo/reference/workspaces.html#workspaces
-    for (key, ws_key) in [
-        ("package", "package"),
-        ("dependencies", "dependencies"),
-        ("dev-dependencies", "dependencies"),
-        ("build-dependencies", "dependencies"),
+    for (key, ws_key, inherit) in [
+        ("package", "package", false),
+        ("dependencies", "dependencies", false),
+        ("dev-dependencies", "dependencies", false),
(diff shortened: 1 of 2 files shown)


pkgs/crane-utils/src/bin/crane-resolve-workspace-inheritance.rs
```diff
@@ -149,7 +173,9 @@ where
         // - the workspace root doesn't have this key
         let (t, root_val) = match try_as_table_like_mut(&mut *v).zip(root.get(&k)) {
             Some((t, root_val)) => (t, root_val),
-            _ => return,
+            _ => {
+                return;
+            }
```
rvolosatovs, line 178: this should not have happened

pkgs/crane-utils/src/bin/crane-resolve-workspace-inheritance.rs
```diff
@@ -420,6 +452,9 @@ mod tests {
             [target.'cfg(unix)'.dependencies]
             unix = { version = "unix-vers" , features = ["some"] }
 
+            [lints.rust]
```
rvolosatovs, line 455: I'm not sure what's up with the placement of these, by the way, but apparently this is deterministic

ipetkov: Not sure either but as long as its deterministic it's fine by me!

ipetkov:
Thanks for fixing this! I peeked at the cargo sources, looks like it doesn't support inheritance and overrides (you can either inherit the workspace lints, or you have to inherit nothing) so this approach looks good!

PR #501, opened 2024-01-26
fix: add #[allow(unused_extern_crates)] to dummy.rs

kallisti-dev:
Closes #500

Issue #500, opened 2024-01-26
default dummySrc fails to build when workspace uses #[deny(unused_extern_crates)]


kallisti-dev:
When attempting to build a workspace with this config in  `Cargo.toml`:

```toml
[workspace.lints.rust]
rust_2018_idioms = "deny"
```

`craneLib.buildDepsOnly` fails with the following error

```
       >  --> /nix/store/xf8kaz2alipnp5qzaysc6kg1cp03z3z2-dummy.rs:5:1
       >   |
       > 5 | extern crate core;
       >   | ^^^^^^^^^^^^^^^^^^ help: remove it
       >   |
       >   = note: `-D unused-extern-crates` implied by `-D rust-2018-idioms`
       >   = help: to override `-D rust-2018-idioms` add `#[allow(unused_extern_crates)]`
```

Seems to be a simple fix: just add `#[allow(unused_extern_crates)]` to this declaration. I'll work on a PR and test locally to see if there are any other conflicts with my workspace lint settings.

kallisti-dev:
I've opened a PR for this change. After updating my local flake inputs to point to the PR revision the error was resolved and workspace was able to build.

PR #506, opened 2024-01-28
test: strict workspace lint levels

kallisti-dev:
As a follow up to #500 and #501, here's a test case that will attempt to build a workspace with all of rustc's lint groups set to the most strict level possible that will still build with the default mkDummySrc

Lints marked as "warn" in the Cargo.toml indicate that those lints currently break the default mkDummySrc when set to "deny" or higher. 

Lints marked as "deny" indicate that they are being explicitly allowed by the default mkDummySrc and thus cannot be set to "forbid" (because forbid prevents overriding)


checks/workspace-lints/Cargo.toml
```diff
@@ -0,0 +1,202 @@
+[workspace]
+members = ["./member/foo"]
+
+[workspace.lints.rust]
```
ipetkov, line 4: *Question*: would it be equivalent (and perhaps simpler) to set `warnings = "deny"` rather than exhaustively listing out all possible lints here?

Happy to strive to make the dummified sources compile with no warnings! Though there wouldn't be anything we can do if the workspace decides to `forbid` any particular lint which we rely on disabling (and if the project really wants that they should consider patching their `[workspace.lints]` tables when using `buildDepsOnly`)

kallisti-dev: > _Question_: would it be equivalent (and perhaps simpler) to set `warnings = "deny"` rather than exhaustively listing out all possible lints here?

That was the plan initially, but then I wanted to test "forbid" all, which you apparently can't do because of [this lint rule](https://doc.rust-lang.org/rustc/lints/listing/warn-by-default.html#forbidden-lint-groups)

There will always need to be some exhaustive listing, because not every lint rule is represented as a lint group. That being said, using the lint groups is a bit more future-proof in that you don't have to worry about new warning lints being added in the future. Another possibility is to use `rustc -W help` which will list out all the lints. There's potential for autogeneration there perhaps. 

You could use `warnings = "deny"` if you're okay with only testing for "deny" and not "forbid", which I think is reasonable since you should know exactly which lints dummy.rs uses since you're explicitly allowing them. My thought was to have this test represent the strictest possible lint configuration that will build with the default dummy code. 

> Happy to strive to make the dummified sources compile with no warnings! Though there wouldn't be anything we can do if the workspace decides to `forbid` any particular lint which we rely on disabling (and if the project really wants that they should consider patching their `[workspace.lints]` tables when using `buildDepsOnly`)

One way to get around some "forbid" rules is to more aggressively use `cfg` to ensure there's no dead code. However I think you will always have an issue with `unused-crate-dependencies` because of the way that the dummy code works (the generated bin/main.rs will not reference the outside crate). 

Agree that the best fix here is to simply patch out the [workspace.lints] if it's an issue. I am wondering if there would be any issue with doing this in the default dummy cargo.toml itself.

kallisti-dev: If cleaning out the [workspace.lints] section isn't an option, then the following code changes I think would allow most lints to pass:

* add documentation on every function and crate/module. allows `missing-docs = "forbid"` to pass.
* add `#![allow(unused-crate-dependencies)]` to the generated bin/main.rs. allows `unused-crate-dependencies = "deny"` to pass.
* place `extern crate foo` and the panic handler inside `cfg` to prevent them from being unused. allows  `dead-code = "forbid"` and `unused-extern-crate = "forbid"` to pass.

I was messing around with the dummy code to make the above changes, but I was still getting a mysterious complaint about `allow(dead_code)` being used with `dead-code = "forbid"` even though I couldn't locate where it was still being used. This happened even after I used `git commit -a --amend` so it wasn't because of dirty git repo.  Perhaps I am missing something about how the dummy code works or how Nix works here.

with the above changes you should be able to pass everything except `unused-crate-dependencies = "forbid"`

ipetkov: > Agree that the best fix here is to simply patch out the [workspace.lints] if it's an issue. I am wondering if there would be any issue with doing this in the default dummy cargo.toml itself

I have a lingering feeling that changing the `[workspace.lints]` definition ends up rebuilding dependencies (I think the lints might be getting captured as rustc args some how?) so stripping it out might lead to artifacts not being appropriately cached... Though if I'm wrong on this it would be super simple to strip out in `cleanCargoToml`

kallisti-dev: > > Agree that the best fix here is to simply patch out the [workspace.lints] if it's an issue. I am wondering if there would be any issue with doing this in the default dummy cargo.toml itself
> 
> I have a lingering feeling that changing the `[workspace.lints]` definition ends up rebuilding dependencies (I think the lints might be getting captured as rustc args some how?) so stripping it out might lead to artifacts not being appropriately cached... Though if I'm wrong on this it would be super simple to strip out in `cleanCargoToml`

Yes that seems to be correct. The error messages I received from rustc indicated that the "forbid" rules were passed via command line arguments. But when compiling crate dependencies, cargo sets all lint rules to "allow"  according to [this documentation](https://doc.rust-lang.org/rustc/lints/levels.html#capping-lints). 

> This feature is used heavily by Cargo; it will pass --cap-lints allow when compiling your dependencies, so that if they have any warnings, they do not pollute the output of your build.

I suspect that means that a recompilation would be avoided, but it's hard to say without testing it.

ipetkov: > I have a lingering feeling that changing the `[workspace.lints]` definition ends up rebuilding dependencies (I think the lints might be getting captured as rustc args some how?)

Did some testing, and I was mistaken here. Changing any of the `[lints]` tables will result in rebuilding the _workspace_ crates, but will not affect dependency crates. This means we can strip them out of the `Cargo.toml` entirely when building dummy sources (since we'd have to rebuild the workspace crates with their real sources anyway there's no issue here).

I've opened https://github.com/ipetkov/crane/pull/516 to start doing that cleaning, which also gives us a nice benefit that any workspace/package lint definitions should not apply to the dummified sources!

kallisti-dev:
Nice! you shouldn't need this test case if that's the case. Though you might want to come up with an easier maintain version of this test just to verify that lints don't interfere.

ipetkov:
Yes I think we can get away with a simple test which sets `--deny warnings` and that should be good enough to catch any unexpected lints!

PR #533, opened 2024-02-26
crane-utils: migrate from toml_edit to toml

nrabulinski:
## Motivation
Currently `toml_edit` is used to resolve `Cargo.toml`s from workspace crates. This was fine but I recently ran into https://github.com/toml-rs/toml/issues/691 which made it impossible for me to compile my project. I thought of trying to fix `toml_edit`, but then I thought to myself that realistically there's very little gain in trying to preserve the comments and order in `Cargo.toml` files which the user won't even see unless, as it was in my case, there's a bug and cargo fails while reading it.

It turned out to be very easy to replace `toml_edit` with `toml`, while also making the code shorter, and (in my opinion) a bit cleaner.

## Checklist
- [x] updated `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   the lint definitions change, and it avoids issues with failing to build
   dummified sources which might have violated a lint marked as `deny` or
   `forbid`
+* `crane-resolve-workspace-inheritance` no longer preserves comments or ordering
 
 ### Fixed
 * Fixed an edge case with inheriting workspace dependencies where the workspace
--- a/pkgs/crane-utils/Cargo.lock
+++ b/pkgs/crane-utils/Cargo.lock
@@ -16,7 +16,7 @@ dependencies = [
  "pretty_assertions",
  "serde",
  "serde_json",
- "toml_edit",
+ "toml",
 ]
 
 [[package]]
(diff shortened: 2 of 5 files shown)


dpc:
There unfortunately seem to be packages do do use comments for stuff https://github.com/ipetkov/crane/issues/394

nrabulinski:
Oh, i did not see that, thanks for pointing that out @dpc. Although IMO there's something wrong with `document-features` if it's supposed to only affect documentation and yet it causes the compilation to fail when there's a missing comment in Cargo.toml

ipetkov:
Thanks for the PR @nrabulinski! Like @dpc mentioned we switched from `toml` to `toml_edit` to accommodate things like `document-features`. It's pretty unfortunate to be stuck waiting on a fix for an upstream bug, but if I had to pick one use case, I'd rather support `document-features` (it's hopefully "easier" to tweak/patch a Cargo.toml to move/remove unnecessary comments IMO)

I'm thinking another solution would be to expose a more-first-class way of customizing fetching git checkouts (namely adding the ability to insert custom patch hooks for example). Then affected projects can more easily tweak downstream Cargo.tomls without having to switch to their own patched branch

ipetkov:
Thanks again for the PR @nrabulinski ! It looks like https://github.com/ipetkov/crane/pull/583 has a workaround for this by ignoring (stripping) all comments when doing workspace inheritance until https://github.com/toml-rs/toml/issues/691 is fixed. Please take a look and see if that resolves your issue!

In the mean time, I'm going to close this PR in favor of https://github.com/toml-rs/toml/issues/691 since we don't want to regress on supporting comments on feature definitions

PR #540, opened 2024-03-04
fix: allow registries in vendorMultipleCargoDeps as documented

greaka:
## Motivation

The docs state that you can use `registries` in `vendorMultipleCargoDeps`. The code does support that. However, it was missed that you need to either allow `registries` directly as a keyword in the function parameters or you need to add a `, ...`. I don't know which version you would prefer, so I simply went with the more strict version for now. Let me know what changes I need to include or feel free to commit to my branch.

## Checklist

- [ ] updated `CHANGELOG.md`


ipetkov (approved):
Looks great, thanks for fixing this!

PR #546, opened 2024-03-07
vendorMultipleCargoDeps: use backwards compatible groupBy (#545)

### LINKED ISSUE #545: builtins.groupBy is not backwards compatible with nix versions before 2.5
The use of `builtins.groupBy` in https://github.com/ipetkov/crane/blob/81c393c776d5379c030607866afef6406ca1be57/lib/vendorMultipleCargoDeps.nix#L11-L15 breaks compatibility with nix versions prior to 2.5.

I suggest switching to `lib.groupBy` (which is already used in other places) to keep backwards compatibility.

Telefragged:
Fixes #545 

## Motivation
Using crane with nix versions prior to 2.5 (2.3 specifically in our case)

## Checklist

- [ ] updated `CHANGELOG.md`

### DIFF
--- a/lib/vendorMultipleCargoDeps.nix
+++ b/lib/vendorMultipleCargoDeps.nix
@@ -12,12 +12,12 @@ let
     attrNames
     attrValues
     fromTOML
-    groupBy
     readFile;
 
   inherit (lib)
     concatMapStrings
-    escapeShellArg;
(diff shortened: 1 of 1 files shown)


ipetkov (approved):
LGTM thanks for addressing this!

PR #548, opened 2024-03-12
crateRegistries: use crates.io's CDN to download sources

ipetkov:
## Motivation
https://blog.rust-lang.org/2024/03/11/crates-io-download-changes.html

## Checklist

- [x] updated `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Changed
+* Sources are now fetched [crates.io's
+  CDN](https://blog.rust-lang.org/2024/03/11/crates-io-download-changes.html),
+  following cargo's (new) default behavior.
+
 ### Fixed:
 * `vendorMultipleCargoDeps` correctly lists `registries` as an optional
   parameter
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -31,7 +31,9 @@ jobs:
       matrix:
         include:
             # Latest and greatest release of Nix
-          - install_url: https://nixos.org/nix/install
+          #- install_url: https://nixos.org/nix/install
+          # extra-tests/alt-store/test.sh is failing with 2.21.0
+          - install_url: https://releases.nixos.org/nix/nix-2.20.5/install
```
dpc, line 36: Confused - is this just a workaround for something? (that security issue?)

ipetkov: Yes our [`extra-test/alt-store`](https://github.com/ipetkov/crane/blob/10484f86201bb94bd61ecc5335b1496794fedb78/extra-tests/alt-store/test.sh#L8) suddenly started failing, though a new version of Nix was released as the latest version which must have some kind of regression given that the tests passed when I pinned it to the previous one

ipetkov: Going to disable this test in https://github.com/ipetkov/crane/pull/556

Issue #527, opened 2024-02-20
I think crane mangles my cargo.toml


pingiun:
My situation:

I have made modifications to sqlx, they live in [my branch](https://github.com/jelle-bigbridge/sqlx/tree/add-local-infile). sqlx is a workspace project

In my own project, I depend on sqlx via git. My own project is also a workspace.

I get this error when building my project

<details>
<summary>Nix build logs</summary>

```
warning: The interpretation of store paths arguments ending in `.drv` recently changed. If this command is now failing try again with '/nix/store/p2c53q2h2419ygdgy4snjrf0klrrsm2h-bibs-server-deps-0.1.0.drv^*'
https://cache.nixos.org/drjna8cp1m2bjphqkra58xpjdqb6s4k1.narinfo
cargoArtifacts not set, will not reuse any cargo artifacts
Running phase: unpackPhase
unpacking source archive /nix/store/21i4hbdcj0kw84b8i8ryraqgfpz1b0kx-source
source root is source
Running phase: patchPhase
Executing configureCargoCommonVars
Running phase: updateAutotoolsGnuConfigScriptsPhase
Running phase: configurePhase
will append /build/source/.cargo-home/config.toml with contents of /nix/store/v1gwn492qw6x8pidrgmnln72a0zrnixm-vendor-cargo-deps/config.toml
default configurePhase, nothing to do
Running phase: buildPhase
++ command cargo --version
cargo 1.75.0 (1d8b05cdd 2023-11-20)
++ command cargo check --profile release-server --package bibs-server --all-targets
error: failed to get `sqlx` as a dependency of package `bibs-server v0.1.0 (/build/source/bibs-server)`

Caused by:
  failed to load source for dependency `sqlx`

Caused by:
  Unable to update https://github.com/jelle-bigbridge/sqlx.git?branch=add-local-infile#6a345589

Caused by:
  failed to update replaced source https://github.com/jelle-bigbridge/sqlx.git?branch=add-local-infile#6a345589

Caused by:
  failed to parse manifest at `/nix/store/v1gwn492qw6x8pidrgmnln72a0zrnixm-vendor-cargo-deps/e4efc4d96eed4dc366eaad252dc8b86b89b113ce9da78099f9fd5f7aba534018/sqlx-mysql-0.7.3/Cargo.toml`

Caused by:
  TOML parse error at line 74, column 14
     |
  74 | [dependencies.
     |              ^
  invalid table header
  expected `.`, `]`
```
</details>

[...]

PR #583, opened 2024-04-11
Fix bug with comment for inline table key

i1i1:
## Motivation
If workspace features dependency and workspace crate uses it, it is always gets merged as regular table (not inline one) so comments ends up in the following position:
```
[
            # arbitrary
            dependencies.arbitrary ]
```

Take a look at the test for more info.

Fixes https://github.com/ipetkov/crane/issues/527

## Checklist

- [x] updated `CHANGELOG.md`


ipetkov (approved):
Thanks for addressing this, totally didn't realize `toml_edit` supported clearing comments like this! It also appears to resolve #527

PR #587, opened 2024-04-14
mkDummySrc: ignore clippy warnings in dummyrs

szlend:
## Motivation

I'd like to be able to run `cargo clippy` in `buildDepsOnly` without it exiting early due to warnings/errors in the dummy file. The PR adds `#![allow(clippy::all)]` to the dummy file.

Why?

It can be tricky to get optimal cargo artifact cache reuse between a custom `mkCargoDerivation` and `buildDepsOnly`.  Especially when you're trying to hide the low level crane/nix details and expose a simple caching interface for CI.

What I've ended up doing is to simply run the exact same cargo command first in `buildDepsOnly` and then `mkCargoDerivation`. This guarantees that everything will be built the exact same way without having to keep all the flags (`--workspace`, `--all-targets`, `--features`, etc.) in sync between the `cargoArtifacts` and the final derivation.

## Checklist

- [ ] updated `CHANGELOG.md`

### DIFF
--- a/checks/clippy/default.nix
+++ b/checks/clippy/default.nix
@@ -1,6 +1,8 @@
 { buildDepsOnly
 , cargoClippy
+, crateNameFromCargoToml
 , linkFarmFromDrvs
+, mkDummySrc
 }:
 
 let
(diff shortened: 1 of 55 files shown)


ipetkov (approved):
Thanks for fixing this! I've added a test case to prevent future regressions

Issue #572, opened 2024-04-01
issues with libring (undefined reference to ring_core_0_17_8_....) while cross compiling to windows 


drishal:
On using an example flake.nix in the repo for cross compilation to windows (https://github.com/ipetkov/crane/blob/master/examples/cross-windows/flake.nix) I am getting these kind of errors during compilation:
```
 /nix/store/f03xyhj7d30cib79lmiw1xvp6xg2qbza-x86_64-w64-mingw32-binutils-2.41/bin/x86_64-w64-mingw32-ld: agent_code/target/x86_64-pc-windows-gnu/release/deps/libring-07cdff39a14e102f.rlib(ring-07cdff39a14e102f.ring.cd26c19d1cffd1c7-cgu.02.rcgu.o):ring.cd26c19d1cffd:(.text+0xaf0): undefined reference to `ring_core_0_17_8_LIMB_shr'
          /nix/store/f03xyhj7d30cib79lmiw1xvp6xg2qbza-x86_64-w64-mingw32-binutils-2.41/bin/x86_64-w64-mingw32-ld: agent_code/target/x86_64-pc-windows-gnu/release/deps/libring-07cdff39a14e102f.rlib(ring-07cdff39a14e102f.ring.cd26c19d1cffd1c7-cgu.02.rcgu.o):ring.cd26c19d1cffd:(.text+0xb7e): undefined reference to `ring_core_0_17_8_LIMBS_less_than'
          /nix/store/f03xyhj7d30cib79lmiw1xvp6xg2qbza-x86_64-w64-mingw32-binutils-2.41/bin/x86_64-w64-mingw32-ld: agent_code/target/x86_64-pc-windows-gnu/release/deps/libring-07cdff39a14e102f.rlib(ring-07cdff39a14e102f.ring.cd26c19d1cffd1c7-cgu.02.rcgu.o):ring.cd26c19d1cffd:(.text+0xb96): undefined reference to `ring_core_0_17_8_LIMBS_are_zero'
[...]

PR #588, opened 2024-04-16
examples/cross-windows: fix issues related to openssl and libring

drishal:
## Motivation
fixes https://github.com/ipetkov/crane/issues/572


ipetkov (approved):
Can't actually test this myself (tried adding `ring` as a dependency but it didn't give me any build errors) but it looks sensible enough

PR #621, opened 2024-05-21
Parallelize removeReferencesToVendoredSources

### LINKED ISSUE #260: Consider adding an API to vendor crates from multiple Cargo.lock files at once
Hello,
Crane is a fantastic addition to my Rust building workflow, however when trying to build a project with `-Zbuild-std`, I run into predictable name resolution errors, since Cargo lacks internet access.

Is there a currently supported way to properly build the `std` crate (or in this case, `core` and `alloc`) under Crane?

### LINKED ISSUE #620: Parallelize removeReferencesToVendoredSourcesHook
I've got a Trunk project which uses [`stdweb`](https://github.com/koute/stdweb) for some async stuff in Wasm. For whatever reason, `stdweb` generates `inline1.js`, `inline2.js`, ..., all the way up to `inline936.js`. Each of these files needs to be processed by `removeReferencesToVendoredSourcesHook`, but they all get processed one by one. Each of them takes about 5 seconds, so the whole build can take over an hour and a half.

The code in `lib/setupHooks/removeReferencesToVendoredSourcesHook.sh` should be changed to go through all the files in `${installLocation}` in parallel, assuming this is possible with Bash. I believe this could be done with GNU Parallel, but I'm not an expert.

I could probably remove the `stdweb` dependency from my project, but I think parallelizing this hook in crane will speed up a lot of builds for other people as well.

DoctorDalek1963:
## Motivation
Closes #620 by using GNU `parallel` to remove references to vendored sources in multiple files at once.

## Checklist
- [X] updated `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Added
+* `removeReferencesToVendoredSourcesHook` now processes multiple files in
+  parallel, speeding up builds
+
 ## [0.17.1] - 2024-05-19
 
 ### Fixed
--- a/lib/setupHooks/removeReferencesToVendoredSources.nix
+++ b/lib/setupHooks/removeReferencesToVendoredSources.nix
@@ -2,6 +2,7 @@
 , makeSetupHook
 , pkgsBuildBuild
 , stdenv
+, parallel
 }:
 
 let
@@ -10,6 +11,7 @@ in
 makeSetupHook
 {
   name = "removeReferencesToVendoredSourcesHook";
+  propagatedBuildInputs = [parallel];
   substitutions = {
     storeDir = builtins.storeDir;
     sourceSigningUtils = lib.optionalString darwinCodeSign ''
--- a/lib/setupHooks/removeReferencesToVendoredSourcesHook.sh
+++ b/lib/setupHooks/removeReferencesToVendoredSourcesHook.sh
@@ -4,8 +4,10 @@ removeReferencesToVendoredSources() {
   local installLocation="${1:-${out:?not defined}}"
   local vendoredDir="${2:-${cargoVendorDir:?not defined}}"
 
-  local installedFile
-  while read installedFile; do
+  removeReferencesInOneFile() {
+    local vendoredDir="${1}"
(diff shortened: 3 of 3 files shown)


DoctorDalek1963:
It's worth noting that this change can make builds non-reproducible. In my case, the only non-reproducible bit is a list of JS files in the `index.html`. It's the same list, but the elements are in different orders.

Non-reproducibility can be really bad for some builds, so this change should probably be opt-in. That would mean the behavior of the hook should change based on something in `mkCargoDerivation`. What's the best way of doing that?

dpc:
Why would it be non-repro? The end result should always be the same, no?

DoctorDalek1963:
Oops, turns out this was just the project I was testing with. It seems like the non-reproducibility is because of the `stdweb` crate and nothing to do with this PR. I've since tested this PR with a different project and the result was reproducible both with and without this parallelism.

dpc:
> It seems like the non-reproducibility is because of the `stdweb` crate and nothing to do with this PR

You think you can follow through with them to address it? Would be great if Rust ecosystem was aware that byte-to-byte reproducibility is important to many people. :)

DoctorDalek1963:
I'd love to, but [`stdweb`](https://github.com/koute/stdweb) has received no new commits in over 5 years, and all the PRs are totally inactive, so it's probably better for anyone using it to switch to something different.

dpc:
> Just chiming in to mention it's not just stdweb that has this issue with trunk builds: My assets/ dir containing some 100 font and png files gets included

Would it make sense to have some hook/filter attribute to disable for bulk stuff that people sometimes have and don't care if it has references, or it possibly can't and there's no point of tryint go fix it up?

antifuchs:
> Would it make sense to have some hook/filter attribute to disable for bulk stuff[...]?

Yes, I was hoping to find a setting for that! But running it in parallel would already be a huge improvement: I think a setting would be a good next step after parallelization lands.

ipetkov:
Thanks again for the PR @DoctorDalek1963 ! I'm going to close this in favor of https://github.com/ipetkov/crane/pull/625 which is able to achieve a really good speedup here without needing to use `parallel` (more context can be found here: https://github.com/ipetkov/crane/issues/620#issuecomment-2131570054 )

I'm going to close this for now, but feel free to reopen if you think there are other improvements to be had!

PR #625, opened 2024-05-25
removeReferencesToVendoredSources: cache vendorDir crawling

ipetkov:
## Motivation
There is no reason to repeatedly crawl the vendor directories for each installed file, we can easily cache the results once up front.

Fixes https://github.com/ipetkov/crane/issues/620

## Checklist

- [x] updated `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Fixed
+* `removeReferencesToVendoredSources` has been optimzed to search for source
+  references only once. For derivations which install many files, this phase can
+  run up to 99% faster than before.
+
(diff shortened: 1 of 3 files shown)


9999years:
I think `xargs --no-run-if-empty signIfRequired` is introducing an error to my build because `signIfRequired` is a shell function and not an executable, so `xargs` can't find it:

https://github.com/NixOS/nixpkgs/blob/7187b3ee8e2687106e8982b02d2890689471579e/pkgs/os-specific/darwin/signing-utils/utils.sh#L38-L43

```
xargs: signIfRequired: No such file or directory
```

Maybe we can work around by doing something like this:

https://github.com/NixOS/nixpkgs/blob/7187b3ee8e2687106e8982b02d2890689471579e/pkgs/os-specific/darwin/signing-utils/auto-sign-hook.sh#L19-L21

PR #631, opened 2024-05-29
Fix `xargs signIfRequired`

9999years:
## Motivation

Fixes a bug introduced with #625.

`signIfRequired` is a shell function, not an executable, so we can't call it with `xargs`.

https://github.com/NixOS/nixpkgs/blob/7187b3ee8e2687106e8982b02d2890689471579e/pkgs/os-specific/darwin/signing-utils/utils.sh#L38-L43

```
xargs: signIfRequired: No such file or directory
```

Uses this approach:

https://github.com/NixOS/nixpkgs/blob/7187b3ee8e2687106e8982b02d2890689471579e/pkgs/os-specific/darwin/signing-utils/auto-sign-hook.sh#L19-L21

Tested locally with my build.

## Checklist

- [ ] updated `CHANGELOG.md`

### DIFF
--- a/lib/setupHooks/removeReferencesToVendoredSources.nix
+++ b/lib/setupHooks/removeReferencesToVendoredSources.nix
@@ -22,10 +22,10 @@ makeSetupHook
         (
           exec 3>&1
           echo signing files:
-          find "''${installLocation}" -type f |
-            sort |
-            tee -a /dev/fd/3 |
-            xargs --no-run-if-empty signIfRequired
+          while IFS= read -r -d $'\0' file; do
+            echo "signing: $file"
+            signIfRequired "$file"
+          done < <(find "''${installLocation}" -type f -print0)
           echo signing done
         )
       fi
--- a/lib/setupHooks/removeReferencesToVendoredSources.nix
+++ b/lib/setupHooks/removeReferencesToVendoredSources.nix
@@ -19,15 +19,12 @@ makeSetupHook
       if [ -n "''${doNotSign-}" ]; then
         echo "not signing as requested";
       else
-        (
-          exec 3>&1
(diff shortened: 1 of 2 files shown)


lib/setupHooks/removeReferencesToVendoredSources.nix
```diff
@@ -22,10 +22,10 @@ makeSetupHook
...
           exec 3>&1
           echo signing files:
-          find "''${installLocation}" -type f |
-            sort |
-            tee -a /dev/fd/3 |
-            xargs --no-run-if-empty signIfRequired
+          while IFS= read -r -d $'\0' file; do
+            echo "signing: $file"
+            signIfRequired "$file"
+          done < <(find "''${installLocation}" -type f -print0)
```
9999years, line 28: Note: Files are no longer sorted, but that's probably fine?

dpc: Does it break determinism in any way?

9999years: It shouldn't, because all the same paths get signed, unless you mean determinism of log output?

dpc: Just double checking. The determinism of log output doesn't matter, AFAIK.

ipetkov: I originally added the sort in there just to keep stuff deterministic in the logs, but it's fine to drop (we can always revisit if there is something more to debug)

ipetkov (approved):
Thanks for fixing this! Sorry for the breakage, didn't have the ability to test it out myself :sweat_smile:

Issue #628, opened 2024-05-28
overrideToolchain does not correctly splice cargo when vendoring git deps


matthiasbeyer:
**Describe the bug**

I have a repository with a cargo workspace that needs to import one dependency via `rev = "<commit hash>"`. Unfortunately, crane does not handle that as expected.

**Reproduction**

I managed to get this issue reproduced even without a workspace here, with the `bevy` repository as dependency which is also the same dependency I am using in my non-minimal example, with the same commit hash actually. Unfortunately I cannot share that actual repository, as it is closed source for now.

Here's the minimal reproduction: https://github.com/matthiasbeyer/test-crane-gitrev

---

The output of a `nix build` call on that repository:

```
error: builder for '/nix/store/mn1rqjzfjlw9zqrlag17ikjr5na21iqi-test-crane-gitrev-deps-0.1.0.drv' failed with exit code 101;                                                                                        
       last 10 log lines:                                                                                                                                                                                           
       > Running phase: configurePhase                                                                                                                                                                              
       > will append /build/source/.cargo-home/config.toml with contents of /nix/store/b7fxdayj4kjhca5szrl4ghpnzgwk600r-vendor-cargo-deps/config.toml                                                               
       > default configurePhase, nothing to do                                                                                                                                                                      
       > Running phase: buildPhase                                                                                                                                                                                  
       > ++ command cargo --version                                                                                                                                                                                 
       > cargo 1.78.0 (54d8815d0 2024-03-26)                                                                                                                                                                        
       > ++ command cargo check --release --locked --all-targets                                                                                                                                                    
       > error: no matching package named `bevy_reflect` found                                                                                                                                                      
       > location searched: https://github.com/bevyengine/bevy.git?rev=6d25545c516b2fb7698334c428b9cefdee12281e#6d25545c                                                                                            
       > required by package `test-crane-gitrev v0.1.0 (/build/source)`                                                                                                                                             
       For full logs, run 'nix-store -l /nix/store/mn1rqjzfjlw9zqrlag17ikjr5na21iqi-test-crane-gitrev-deps-0.1.0.drv'.                                                                                              
error: 1 dependencies of derivation '/nix/store/7q7f3k60bslx9g3pwq0bpp0qhh3h9jjy-test-crane-gitrev-0.1.0.drv' failed to build   
```

ipetkov:
Hey @matthiasbeyer thanks for the bug report and the reproduction!

[...]

Issue #551, opened 2024-03-17
Not possible to override crane-utils or its dependencies


fracek:
Hello, I need to compile my project using an older `nixpkgs` (23.05). The issue I'm facing is that it tries to compile `crane-utils` using nix `rustPlatform` which uses Cargo 1.69 but the minimum required version to compile `crane-utils` is 1.70.

I haven't figured out how to override the `rustPlatform` used to build `crane-utils`. If I override `crane-utils` in an overlay it's not used by crane anyway.

[I made a minimal repository to reproduce the issue here](https://github.com/fracek/crane-with-nix-23-05)

ipetkov:
Hi @fracek thanks for the report! At this time we don't have a good way to override/overlay crane-utils: https://github.com/ipetkov/crane/blob/f1dabc4abaa6e1ef4ef92d046aced96d1e94dc14/lib/downloadCargoPackageFromGit.nix#L12

In the mean time you can get unblocked by either using `crane-0.15.1` or using the `23.11` release branch. `23.05` is EOL and we dropped support for it in `crane-0.16.0`

ipetkov:
Overriding `craneUtils` will become possible once https://github.com/ipetkov/crane/pull/652 lands!

PR #634, opened 2024-06-03
lib: use makeScopeWithSplicing' instead of makeScope

ipetkov:
## Motivation

Use a spliced scope that better allows for (more correctly) controlling what tools are used where when cross compiling.

Fixes https://github.com/ipetkov/crane/issues/628
Possibly related to https://github.com/ipetkov/crane/issues/551, though I still could not resolve that one yet

## Checklist

- [ ] updated `CHANGELOG.md`

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,5 +1,13 @@
 { pkgs ? import <nixpkgs> { } }:
 
 import ./lib {
-  inherit (pkgs) lib newScope;
+  inherit (pkgs) lib makeScopeWithSplicing';
+  otherSplices = {
+    selfBuildBuild = pkgs.pkgsBuildBuild;
+    selfBuildHost = pkgs.pkgsBuildHost;
+    selfBuildTarget = pkgs.pkgsBuildTarget;
+    selfHostHost = pkgs.pkgsHostHost;
+    selfHostTarget = pkgs.pkgsHostTarget;
+    selfTargetTarget = pkgs.pkgsTargetTarget;
+  };
 }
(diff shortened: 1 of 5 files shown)


default.nix
```diff
@@ -1,5 +1,13 @@
 { pkgs ? import <nixpkgs> { } }:
 
 import ./lib {
-  inherit (pkgs) lib newScope;
+  inherit (pkgs) lib makeScopeWithSplicing';
+  otherSplices = {
+    selfBuildBuild = pkgs.pkgsBuildBuild;
```
simonzkl, line 6: I don't really understand how this solves the issue. `selfBuildBuild` should be `buildBuild` variants of crane's packages, not the ones from nixpkgs. Splicing `cargo` will just pick `cargo` from `pkgsBuildBuild`, same as before.

simonzkl: Oh, actually I guess it does work if you pass a toolchain overlayed over nixpkgs.

ipetkov: yes this was exactly the problem, using `craneLib.overrideToolchain` would use `.overrideScope` to define a new `cargo` definition, except that wouldn't make it into the spliced packages themselves, so the version from `nixpkgs` was being selected (which was still "wrong" even if spliced "correctly")

szlend: Nope, this doesn't work. `callPackage` will splice `cargo` here as `pkgsBuildBuild.cargo`.

I got a few idea on how to fix this in #648.

Issue #648, opened 2024-06-16
Crane is no longer able to cross-compile with an overridden toolchain


szlend:
**Describe the bug**

After #634 crane is no longer able to cross-compile with an overridden toolchain. Instead it will always use `cargo`/`rustc` from `nixpkgs`.

**Reproduction**

1. Check out https://github.com/szlend/test-crane-gitrev-cross (based on [cross-rust-overlay](https://github.com/ipetkov/crane/tree/master/examples/cross-rust-overlay))
2. Run `nix-tree --derivation .#default`
3. Only `rustc-1.73.0` is in the build closure (expected `rustc-1.79.0`)

Removing `crossSystem` fixes the issue and `rustc-1.79.0` is used again.

szlend:
I think `makeScopeWithSplicing'` in #634 is likely used incorrectly. Iirc `self<host><target>` should refer to a scope where `cargo` and `rustc` are actually variants of the toolchain we passed in and not from `pkgs<host><target>`.

szlend:
I managed to work around this issue like so:

```nix
mkToolchain = p:
  let
    toolchain = p.rust-bin.stable.latest.default;
  in
  {
    cargo = toolchain;
    clippy = toolchain;
    rustc = toolchain;
    rustfmt = toolchain;
  };

otherSplices = {
  selfBuildBuild = mkToolchain pkgs.pkgsBuildBuild;
  selfBuildHost = mkToolchain pkgs.pkgsBuildHost;
  selfBuildTarget = mkToolchain pkgs.pkgsBuildTarget;
  selfHostHost = mkToolchain pkgs.pkgsHostHost;
  selfHostTarget = mkToolchain pkgs.pkgsHostTarget;
  selfTargetTarget = { }; # usually empty
};

craneLib = import "${crane}/lib" {
  inherit (pkgs) lib makeScopeWithSplicing';
  inherit otherSplices;
};
```

I'm not sure what the best way to expose this as an API. A few ideas:
1. Let users pass a custom `otherSplices`
2. Make `overrideToolchain` accept a function like: `craneLib.overrideToolchain (p: p.rust-bin.stable.latest.default)` where `p` is a variant of `pkgs<host><target>`. Though this assumes that you overlay the toolchain over nixpkgs.
3. Both?

ipetkov:
[...]

PR #650, opened 2024-06-20
PoC: Cross-compilation support for custom toolchains

szlend:
## Motivation

PoC fix for #648

```nix
craneLib = crane.mkLibForToolchain pkgs (pkgs: pkgs.rust-bin.stable.latest.default);
```

## Checklist

- [ ] updated `CHANGELOG.md`


szlend:
It's probably better to do #652 if we don't want to break the API.

PR #652, opened 2024-06-23
Splice overrideToolchain

szlend:
## Motivation

`overrideToolchain` can now accept a function used to splice the rust toolchain:

```nix
craneLib = (crane.mkLib pkgs).overrideToolchain (p: p.rust-bin.stable.latest.default);
```

~~I don't fully understand why the toolchain needs be offset from `pkgsBuildHost` even when spliced. But this has always been the case.~~

Fixes https://github.com/ipetkov/crane/issues/648

Fixes https://github.com/ipetkov/crane/issues/551

## Checklist

- [x] updated `CHANGELOG.md`

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,14 +1,3 @@
 { pkgs ? import <nixpkgs> { } }:
 
-import ./lib {
-  inherit (pkgs)
-    lib
-    makeScopeWithSplicing'
-    splicePackages
-    pkgsBuildBuild
-    pkgsBuildHost
-    pkgsBuildTarget
-    pkgsHostHost
-    pkgsHostTarget
-    pkgsTargetTarget;
-}
+pkgs.callPackage ./lib { }
--- a/default.nix
+++ b/default.nix
@@ -1,13 +1,14 @@
 { pkgs ? import <nixpkgs> { } }:
 
 import ./lib {
-  inherit (pkgs) lib makeScopeWithSplicing';
-  otherSplices = {
-    selfBuildBuild = pkgs.pkgsBuildBuild;
-    selfBuildHost = pkgs.pkgsBuildHost;
-    selfBuildTarget = pkgs.pkgsBuildTarget;
-    selfHostHost = pkgs.pkgsHostHost;
-    selfHostTarget = pkgs.pkgsHostTarget;
-    selfTargetTarget = pkgs.pkgsTargetTarget;
-  };
+  inherit (pkgs)
+    lib
+    makeScopeWithSplicing'
+    splicePackages
+    pkgsBuildBuild
+    pkgsBuildHost
+    pkgsBuildTarget
+    pkgsHostHost
+    pkgsHostTarget
+    pkgsTargetTarget;
 }
--- a/docs/API.md
+++ b/docs/API.md
@@ -1193,15 +1193,20 @@ build caches. More specifically:
 
 ### `craneLib.overrideToolchain`
 
-`overrideToolchain :: drv -> set`
+`overrideToolchain :: (set -> drv) -> set`
+`overrideToolchain :: drv -> set` (deprecated)
 
 A convenience method to override and use tools (like `cargo`, `clippy`,
 `rustfmt`, `rustc`, etc.) from one specific toolchain. The input should be a
 single derivation which contains all the tools as binaries. For example, this
 can be the output of `oxalica/rust-overlay`.
 
+Note that in order to best support cross compilation, `overrideToolchain` should
+be provided a function (whose argument is a cross-compilation aware version of
+`pkgs`) which constructs the toolchain:
+
 ```nix
-craneLib.overrideToolchain myCustomToolchain
+craneLib.overrideToolchain (p: myCustomToolchainForPkgs p)
 ```
 
 ### `craneLib.path`
--- a/docs/API.md
+++ b/docs/API.md
@@ -1194,7 +1194,7 @@ build caches. More specifically:
 ### `craneLib.overrideToolchain`
 
 `overrideToolchain :: (set -> drv) -> set`
-`overrideToolchain :: drv -> set` (deprecated)
+`overrideToolchain :: drv -> set` (legacy)
 
 A convenience method to override and use tools (like `cargo`, `clippy`,
 `rustfmt`, `rustc`, etc.) from one specific toolchain. The input should be a
--- a/examples/cross-rust-overlay/flake.nix
+++ b/examples/cross-rust-overlay/flake.nix
@@ -31,7 +31,7 @@
           overlays = [ (import rust-overlay) ];
         };
 
-        craneLib = (crane.mkLib pkgs).overrideToolchain (pkgs: pkgs.rust-bin.stable.latest.default);
+        craneLib = (crane.mkLib pkgs).overrideToolchain (p: p.rust-bin.stable.latest.default);
 
         # Note: we have to use the `callPackage` approach here so that Nix
         # can "splice" the packages in such a way that dependencies are
--- a/examples/cross-rust-overlay/flake.nix
+++ b/examples/cross-rust-overlay/flake.nix
@@ -31,11 +31,7 @@
           overlays = [ (import rust-overlay) ];
         };
 
-        mkRustToolchain = pkgs: pkgs.pkgsBuildHost.rust-bin.stable.latest.default.override {
-          targets = [ "aarch64-unknown-linux-gnu" ];
-        };
-
-        craneLib = (crane.mkLib pkgs).overrideToolchain mkRustToolchain;
+        craneLib = (crane.mkLib pkgs).overrideToolchain (pkgs: pkgs.rust-bin.stable.latest.default);
 
         # Note: we have to use the `callPackage` approach here so that Nix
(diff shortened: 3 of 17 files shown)


default.nix
```diff
@@ -1,13 +1,14 @@
...
+  inherit (pkgs)
+    lib
+    makeScopeWithSplicing'
+    splicePackages
+    pkgsBuildBuild
+    pkgsBuildHost
+    pkgsBuildTarget
+    pkgsHostHost
+    pkgsHostTarget
+    pkgsTargetTarget;
```
szlend, line 13: Any reason why this couldn't just be `pkgs.callPackage ./lib {}`?

lib/default.nix
```diff
@@ -9,85 +15,121 @@ let
...
+      installFromCargoBuildLogHook = callPackage ./setupHooks/installFromCargoBuildLog.nix { };
+      mkCargoDerivation = callPackage ./mkCargoDerivation.nix { };
+      mkDummySrc = callPackage ./mkDummySrc.nix { };
 
-  path = callPackage ./path.nix {
-    inherit internalCrateNameForCleanSource;
-  };
+      overrideToolchain = toolchain: self.overrideScope (_final: _prev:
+        let
+          splicedToolchain = spliceToolchain toolchain;
```
szlend, line 88: Manually splice the toolchain using `pkgs.splicePackages` instead of relying on `otherSplices`. I don't think there's any other way if we're using `overrideScope`, because `overrideScope` doesn't update `otherSplices`.

szlend:
Actually, on second testing, it doesn't seem there's any difference to just doing:

```nix
craneLib = (crane.mkLib pkgs).overrideToolchain pkgsBuildHost.rust-bin.stable.latest.default;
```

So it kind of seems like splicing isn't working correctly. Though I did verify that `cargo` passed into `nativeBuildInputs` in various crane derivations has the `__spliced` attrset. So I don't understand what's going on here.

lib/default.nix
```diff
@@ -9,85 +16,118 @@ let
...
+    f = scopeFn;
+    otherSplices = {
+      selfBuildBuild = lib.makeScope pkgsBuildBuild.newScope scopeFn;
+      selfBuildHost = lib.makeScope pkgsBuildHost.newScope scopeFn;
+      selfBuildTarget = lib.makeScope pkgsBuildTarget.newScope scopeFn;
+      selfHostHost = lib.makeScope pkgsHostHost.newScope scopeFn;
+      selfHostTarget = lib.makeScope pkgsHostTarget.newScope scopeFn;
+      selfTargetTarget = lib.optionalAttrs (pkgsTargetTarget?newScope) (lib.makeScope pkgsTargetTarget.newScope scopeFn);
+    };
+    keep = self: lib.optionalAttrs (self?cargo) { inherit (self) cargo clippy rustc rustfmt; };
```
szlend, line 130: This is a bit of a hack to make `overrideScope` work as expected. The `keep` argument tells `makeScopeWithSplicing'` to keep those packages alone (don't splice them), which is fine because we manually splice them in `overrideScope` anyway. This is skipped if `cargo` isn't overridden.

This is kind of brittle so it probably deserves a unit test. Though we don't currently unit test cross-compilation.

szlend:
> I don't fully understand why the toolchain needs be offset from `pkgsBuildHost` even when spliced. But this has always been the case.

I should have seen this as a red flag. This works fine now:

```nix
craneLib = (crane.mkLib pkgs).overrideToolchain (p: p.rust-bin.stable.latest.default);
```

default.nix
ipetkov: No reason, totally fine to use `pkgs.callPacakge` here!

lib/default.nix
ipetkov: > Though we don't currently test cross-compilation.

We do have the cross-rust-overlay example, is that not sufficient? Unless you mean we don't have a test that we didn't break splicing `rustfmt` for example

lib/default.nix
```diff
@@ -9,85 +16,118 @@ let
...
-  };
+      overrideToolchain = toolchainArg: self.overrideScope (_final: _prev:
+        let
+          toolchain =
+            if lib.isFunction toolchainArg
+            then spliceToolchain toolchainArg
+            else toolchainArg;
+          needsSplicing = stdenv.buildPlatform != stdenv.hostPlatform && toolchain?__spliced == false;
+        in
+        lib.warnIf needsSplicing "crane overrideToolchain requires a spliced toolchain when cross-compiling" {
```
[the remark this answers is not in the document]
ipetkov: I think the warning is good to keep! There can be a lot of subtle errors when setting up cross compilation, if we can catch a possible bad situation and at least warn on it it would be helpful to users (plus its costs us very little to have a warning here so I'm all for it)

examples/cross-rust-overlay/flake.nix
```diff
@@ -31,11 +31,7 @@
           overlays = [ (import rust-overlay) ];
         };
 
-        rustToolchain = pkgs.pkgsBuildHost.rust-bin.stable.latest.default.override {
-          targets = [ "aarch64-unknown-linux-gnu" ];
-        };
-
-        craneLib = (crane.mkLib pkgs).overrideToolchain rustToolchain;
+        craneLib = (crane.mkLib pkgs).overrideToolchain (pkgs: pkgs.rust-bin.stable.latest.default);
```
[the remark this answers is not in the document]
ipetkov: Nice this is way cleaner too!

lib/default.nix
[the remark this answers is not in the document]
ipetkov: Though we may want to tweak the warning to hint towards the solution (namely, use a function instead of passing in a value directly), esp since this used to require an explicit value and anyone used to the older format might not understand what changed

ipetkov (commented):
Thanks again for the PR, and for making it backwards compatible! Overall I think it looks good. The hoops we have to jump through to make everything work are no where near as bad as I feared so I'm okay with moving in that direction!

Going to do a little bit of my own testing and move forward from there if it looks fine

lib/default.nix
szlend: Yeah we don't have any tests for splicing. I think we could do this with an eval time test by checking that `__spliced.buildHost` is present on `cargo` (checking only for `__spliced` is not enough from my experience). But not sure where this fits into `crane`.

lib/default.nix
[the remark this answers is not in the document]
ipetkov: Reworded this comment!

docs/API.md
```diff
@@ -1193,15 +1193,20 @@ build caches. More specifically:
 
 ### `craneLib.overrideToolchain`
 
-`overrideToolchain :: drv -> set`
+`overrideToolchain :: (set -> drv) -> set`
+`overrideToolchain :: drv -> set` (deprecated)
```
szlend, line 1197: Do we want to deprecate this? Imo it's perfectly fine to use the old form when not cross-compiling and we do have a warning in place if you try to do this with a cross nixpkgs.

Though I suppose if you don't care (or your toolchain is not in scope of `pkgs`) you can just as well do:
```
overrideToolchain (_: myToolchain)
```

ipetkov: I guess it's not so much "deprecated" as "discouraged"? As in using the function-method definitely works in all cases, while using a direct value only works if you don't care about cross-compiling (though if you suddenly started to care, our warning would catch that).

I'll rename this to `legacy` so it doesn't carry as heavy as a connotation!

lib/downloadCargoPackageFromGit.nix
```diff
@@ -8,8 +9,6 @@ let
   inherit (pkgsBuildBuild)
     fetchgit
     stdenv;
-
-  craneUtils = pkgsBuildBuild.callPackage ../pkgs/crane-utils { };
```
szlend, line 12: Oh this is a nice side benefit. When I spliced crane itself I didn't consider we could actually use this ourselves. I only did it cause it seemed the correct thing to do.

ipetkov: Yeah this was a workaround to force the correct version to be used before splicing was fully sorted out, but now that you came along and fixed things end to end it was time to get rid of it :smile:

lib/default.nix
ipetkov: I've been trying to think of a way to achieve this test but without exposing any cruft to callers, but so far I'm drawing a blank... We can always shove a custom eval script in `./extra-tests` that does whatever dirty hacks it needs to, but I'm also okay with leaving it without a test if we can't really write one in a good way...

ipetkov: Oh worth noting that I DRY'd this bit of logic so I'm less concerned about us forgetting to keep any of `cargo`/`clippy`/`rustc`/`rustfmt` if we ever gain a new one. And if at that point splicing is still broken we probably have a different issue to fix :smile:

szlend: Tbh it's probably fine to go without one. The way the cross-compilation example is set up now (without explicit `pkgsBuildHost` and `targets`), it should fail if anything in the splicing chain goes wrong.

ipetkov: Cool, let's merge this in then, thanks again for the great work!

ipetkov (approved):

PR #665, opened 2024-07-15
Add build-std and quick-start-workspace as flake templates

SebRut:
## Motivation

I'd like to use the workspace template as documented on https://crane.dev/examples/quick-start-workspace.html . Running the provided command fails with `error: flake 'github:ipetkov/crane' does not provide attribute 'templates.quick-start-workspace' or 'quick-start-workspace'`, I think because the example is missing from the flake.

## Checklist

- [ ] updated `CHANGELOG.md`

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -94,6 +94,11 @@
           path = ./examples/quick-start-simple;
         };
 
+        quick-start-workspace = {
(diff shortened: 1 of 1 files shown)


ipetkov (approved):
Good catch, thanks for fixing this!

PR #674, opened 2024-07-30
Add `taploFmt` TOML format function to `mkLib`

### LINKED ISSUE #673: Add `Cargo.toml` formatting check to `checks`

It would be nice to have a function available for checking the formatting of `Cargo.toml`s

Add a function similar to the existing format check, or extend the existing format check to include `Cargo.toml` formatting.
```nix
checks = {
  my-crate-fmt = craneLib.cargoFmt {
    src = myCrateSrc;
  };

  my-crate-toml-fmt = craneLib.tomlFmt {
    src = myCrateSrc;
  };
}
```
I'm using `taplo` but whichever is the most popular would be fine.

If this isn't something `crane` should do for me, I could do this myself using `crane` as inspiration, but I figured I'd check in here first.

eureka-cpu:
## Motivation

Closes #673 

Hello, this change adds a function to `mkLib` for checking formatting of TOML files. TOML files are extremely common, especially among Rust projects, and I find that in all of my use cases of `crane` between projects I am always manually adding checks for the formatting of `Cargo.toml` files, and other files such as `deny.toml`, `rust-toolchain.toml`, `rustfmt.toml` and so on.

I have a working copy of this on a local project, otherwise I wasn't sure how to add tests for it. I'm also sure that the copy-pasted code I took from `cargoFmt.nix` may include some things that don't belong, or may otherwise be going about it the wrong way. Regardless, this is what worked for me, please do instruct me on how to get this to a mergable state, I would greatly appreciate it!

## Checklist

- [x] added tests to verify new behavior (toml format is added to checks so it is already tested in CI)
- [ ] updated `CHANGELOG.md`

### DIFF
--- a/examples/quick-start-workspace/flake.nix
+++ b/examples/quick-start-workspace/flake.nix
@@ -120,6 +120,11 @@
             inherit src;
           };
 
+          my-workspace-toml-fmt = craneLib.taploFmt {
+            inherit src;
+            taploExtraArgs = "--config ./taplo.toml";
+          };
+
           # Audit dependencies
           my-workspace-audit = craneLib.cargoAudit {
             inherit src advisory-db;
--- a/examples/quick-start-workspace/flake.nix
+++ b/examples/quick-start-workspace/flake.nix
@@ -121,7 +121,7 @@
           };
 
           my-workspace-toml-fmt = craneLib.taploFmt {
-            inherit src;
+            src = pkgs.lib.sources.sourceFilesBySuffices src [ ".toml" ];
             taploExtraArgs = "--config ./taplo.toml";
           };
 
--- a/examples/quick-start-workspace/flake.nix
+++ b/examples/quick-start-workspace/flake.nix
@@ -122,7 +122,8 @@
 
           my-workspace-toml-fmt = craneLib.taploFmt {
             src = pkgs.lib.sources.sourceFilesBySuffices src [ ".toml" ];
-            taploExtraArgs = "--config ./taplo.toml";
+            # taplo arguments can be further customized below as needed
+            # taploExtraArgs = "--config ./taplo.toml";
           };
 
           # Audit dependencies
--- a/examples/quick-start/flake.nix
+++ b/examples/quick-start/flake.nix
@@ -93,7 +93,8 @@
 
           my-crate-toml-fmt = craneLib.taploFmt {
             src = pkgs.lib.sources.sourceFilesBySuffices src [ ".toml" ];
-            taploExtraArgs = "--config ./taplo.toml";
+            # taplo arguments can be further customized below as needed
+            # taploExtraArgs = "--config ./taplo.toml";
           };
 
           # Audit dependencies
--- a/examples/quick-start/flake.nix
+++ b/examples/quick-start/flake.nix
@@ -91,6 +91,11 @@
             inherit src;
           };
 
+          my-crate-toml-fmt = craneLib.taploFmt {
+            inherit src;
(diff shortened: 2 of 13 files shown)


lib/taploFmt.nix
```diff
@@ -0,0 +1,24 @@
+{ taplo
+, mkCargoDerivation
+}:
+
+{ taploExtraArgs ? ""
```
ipetkov, line 5: Maybe we can make the default here `"--config ./toplo.toml"` and spare callers from needing to pass it in themselves (but they are welcome to override it completely)

examples/quick-start-workspace/flake.nix
```diff
@@ -120,6 +120,11 @@
             inherit src;
           };
 
+          my-workspace-toml-fmt = craneLib.taploFmt {
+            inherit src;
```
ipetkov, line 124: Might be good to highlight here that we can use a source filter to cut down the number of times the derivation needs to be rebuilt when _other_ source files change.

For example:

```suggestion
            src = pkgs.lib.sources.sourceFilesBySuffices src [ ".toml" ];
```

examples/quick-start/flake.nix
```diff
@@ -91,6 +91,11 @@
             inherit src;
           };
 
+          my-crate-toml-fmt = craneLib.taploFmt {
+            inherit src;
```
ipetkov, line 95: Same comment

```suggestion
            src = pkgs.lib.sources.sourceFilesBySuffices src [ ".toml" ];
```

ipetkov (commented):
Hi @eureka-cpu thanks for the PR!

Although I haven't seen the broader Rust community standardize on a toml formatter, this is is pretty self contained so I'm open to including it!

The last thing we need to do before merging this is updating the API docs to document how `taploFmt` works! You can adapt the existing docs for `cargoFmt` since they're very similar

eureka-cpu:
> Hi @eureka-cpu thanks for the PR!
> 
> 
> 
> Although I haven't seen the broader Rust community standardize on a toml formatter, this is is pretty self contained so I'm open to including it!
> 
> 
> 
> The last thing we need to do before merging this is updating the API docs to document how `taploFmt` works! You can adapt the existing docs for `cargoFmt` since they're very similar

Thanks for the review! I'll have those changes up shortly (:

lib/taploFmt.nix
eureka-cpu: Just now getting back to this. 

The only downside of using `"--config ./taplo.toml"` as the default is that if the user doesn't have that file, it will show an error and then use the default options. Maybe something cool we could do is add a field called `useConfig` which will search the file tree for a `.taplo.toml`/`taplo.toml`(more info on that here https://taplo.tamasfe.dev/configuration/file.html) and if it doesn't find one then it will show a more helpful error instead of passing over the error and continuing, potentially causing the user to misinterpret the check as successfully using their config file.

eureka-cpu: I use the `taploExtraArgs` in the examples to make it more clear to new users where they can find the `taplo.toml` and make adjustments. If you'd like maybe I could also include a link to the configuration documentation above for other options?

eureka-cpu: Gone ahead and added the link to the docs in the API section for the function

ipetkov: I'm good leaving the default args blank here in that case! I also tweaked the examples to comment out the taplo configs with a comment saying it's possible (but not required) to define it!

ipetkov (approved):
Thanks again for the contribution!

PR #676, opened 2024-08-03
Fix deeply nested registries

### LINKED ISSUE #456: Recursive registry vendoring doesn't work
My crate `foo` has a dependency on crate `bar` from custom registry `registry1`, and `bar` is itself dependent on crate `baz` from `registry2`. `registry1` and `registry2` are in fact **two different URLs for the same registry**, one of them is `sparse+http://` and the other one is `git://` (but this doesn't matter in the general case).

Even if I configure Crane with both registries, it only vendors crates from `registry1`. Then it tries to access `registry2` during `cargo check` in another derivation, which of course fails due to network access issues.

How hard do you think it would be to implement proper recursive vendoring?

Is it possible to temporarily work around this issue without modifying the crates in the custom registry?

-----

I tried to put the following config into `.cargo/config.toml`: 
```toml
# registry used to download `bar`
[registries.registry1]
index = "sparse+http://example.com/api/v1/crates/"

# the registry that `bar` wants `baz` from
[source.replacement2]
registry = "git://example.com/index"
replace-with = "registry1"
```

It fixed the issue of trying to access `registry2` (`git://example.com/index`) from sandbox, but it was causing errors like this:
```
error[E0308]: mismatched types
   --> /nix/store/8krp2f7953dmzlg6ds096dafnhyh1yga-vendor-cargo-deps/059aab26819e2c754aab8ea711140537883f6c4c885fb914bc87eaf4194b5795/bar-0.2.3/src/watermelon.rs:454:21
    |
454 |                     redacted,
    |                     ^^^^^^^^ expected `FooBarType<u8>`, found a different `FooBarType<u8>`
    |
    = note: `FooBarType<u8>` and `FooBarType<u8>` have similar names, but are actually distinct types
note: `FooBarType<u8>` is defined in crate `baz`
   --> /nix/store/8krp2f7953dmzlg6ds096dafnhyh1yga-vendor-cargo-deps/059aab26819e2c754aab8ea711140537883f6c4c885fb914bc87eaf4194b5795/baz-0.1.1/src/banana.rs:32:1
    |
32  | pub struct FooBarType<T>(Box<[T]>);
    | ^^^^^^^^^^^^^^^^^^^^^^^^
note: `FooBarType<u8>` is defined in crate `baz`
   --> /nix/store/8krp2f7953dmzlg6ds096dafnhyh1yga-vendor-cargo-deps/059aab26819e2c754aab8ea711140537883f6c4c885fb914bc87eaf4194b5795/baz-0.1.1/src/banana.rs:32:1
    |
32  | pub struct FooBarType<T>(Box<[T]>);
    | ^^^^^^^^^^^^^^^^^^^^^^^^
    = note: perhaps two different versions of crate `baz` are being used?
```

and looking at the output of `cargo check` we see that it was indeed working with two identical copies of `baz` (from both sources, most likely because: `foo -> bar(registry1) -> baz(registry2)`, `foo -> baz(registry1)`, `foo` also depends on `baz`):
```
    Checking baz v0.1.1 (registry `replacement2`)
    Checking baz v0.1.1 (registry `registry1`)
```

Note that it works fine when running `cargo check` with the modified `.cargo/config.toml` outside of Crane:
```
    Checking baz v0.1.1 (registry `registry1`)
```

szlend:
## Motivation

This allows building packages with crates from custom registries that reference other custom registries.

It works by scanning packages from `Cargo.lock` and adding any missing registries to crane's generated `.cargo/config.toml`. This matches Cargo's behavior where the top level project doesn't necessarily have to explicitly define all transitive registries.

Fixes #456 

## Checklist

- [ ] updated `CHANGELOG.md`

### DIFF
--- a/lib/vendorCargoRegistries.nix
+++ b/lib/vendorCargoRegistries.nix
@@ -1,4 +1,5 @@
-{ downloadCargoPackage
+{ crateRegistries
+, downloadCargoPackage
 , lib
 , pkgsBuildBuild
 }:
@@ -23,18 +24,23 @@ let
     concatMapStrings
     concatStrings
     escapeShellArg
+    filterAttrs
     flatten
+    foldl
     groupBy
     hasPrefix
     mapAttrs'
     mapAttrsToList
-    nameValuePair;
+    nameValuePair
+    removePrefix;
 
   inherit (lib.lists) unique;
 
   hash = hashString "sha256";
 
   hasRegistryProtocolPrefix = s: hasPrefix "registry+" s || hasPrefix "sparse+" s;
+
+  removeProtocol = s: removePrefix "registry+" (removePrefix "sparse+" s);
 in
 { cargoConfigs ? [ ]
 , lockPackages
@@ -57,11 +63,17 @@ let
     '') packages}
   '';
 
+  # Registries configured in cargo config
   parsedCargoConfigTomls = map (p: builtins.fromTOML (readFile p)) cargoConfigs;
   allCargoRegistries = flatten (map (c: c.registries or [ ]) parsedCargoConfigTomls);
   allCargoRegistryPairs = flatten (map (mapAttrsToList (name: value: { inherit name value; })) allCargoRegistries);
   allCargoRegistryPairsWithIndex = filter (r: r ? value.index) allCargoRegistryPairs;
   configuredRegistries = mapAttrs (_: map (r: r.value.index)) (groupBy (x: x.name) allCargoRegistryPairsWithIndex);
+
+  # Registries referenced in Cargo.lock that are missing from cargo config
+  existingRegistries = foldl (acc: r: acc // { ${removeProtocol r.value.index} = true; }) { "https://github.com/rust-lang/crates.io-index" = true; } allCargoRegistryPairsWithIndex;
+  allPackageRegistries = foldl (acc: p: if p ? source then acc // { ${removeProtocol p.source} = [ p.source ]; } else acc) { } lockPackages;
+  missingPackageRegistries = filterAttrs (name: _: !(existingRegistries ? ${name})) allPackageRegistries;
 
   # Append the default crates.io registry, but allow it to be overridden
   registries = {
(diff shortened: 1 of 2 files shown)


szlend:
Not sure what the best way would be to add a regression test. Should we rely on third party registries?

ipetkov (approved):
Thank you so much for fixing this @szlend ! This solution looks great, the [cargo docs on source replacement](https://doc.rust-lang.org/cargo/reference/source-replacement.html) do not cover this at all but using the registry url in place of `[source.${name}]` is definitely the correct thing to do because this is exactly what `cargo vendor` does. I also confirmed that this does fix the build for https://github.com/szlend/crane-reg-test (after adding in `pkg-config` and `openssl` to the inputs that is).

> Not sure what the best way would be to add a regression test. We already rely on alexandria for alt-registry tests, but we'd need another registry. In my reproduction example (https://github.com/szlend/crane-reg-test) I simply forked the alexandria index to have a second copy of the registry.

I do think your reproduction is a great addition to have, but happy to do the same under my own name so you aren't on the hook for not breaking the test suite ;)

ipetkov:
Is there anything you had to do to get cargo to mix and match the registries in the `Cargo.lock` file or did you manually edit one in to reproduce the issue?

ipetkov:
Oh I see you published an actual crate to the upstream index

ipetkov:
Not sure the best way to proceed tbh, seems like we'd need to push more versions of crane test crates to the alexandrie index to untangle but I'm not 100% sure if we want to have to rely on them that way. Perhaps at some point it may be worth looking into standing up a https://kellnr.io instance just for our own testing (maybe even an ephemeral one in a NixOS VM test?), but I definitely don't want to block on that. Happy to merge this as in and fix the issue and we can circle back if we ever have a regression

lib/vendorCargoRegistries.nix
```diff
@@ -1,4 +1,5 @@
-{ downloadCargoPackage
+{ crateRegistries
```
szlend, line 1: Just realized this is no longer needed. I used this before getting registries from `lockPackages`.

ipetkov: Good catch! Fixed in b7f607c2004246d2784f4a58f60a5840066a872b

Issue #675, opened 2024-08-02
Cannot set `cargoExtraArgs` in `craneLib.cargoNextest`


skarmux:
https://github.com/ipetkov/crane/blob/529c1a0b1f29f0d78fa3086b8f6a134c71ef3aaf/lib/cargoNextest.nix#L46-L51

```
$ cargo nextest run -h
--
Build and run tests

Usage: cargo nextest run [OPTIONS] [FILTERS]... [-- <FILTERS_AND_ARGS>...]
[...]
```

As stated in the documentation of the **cargo-nextest** crate, any extra argument/option is to be set behind `$ cargo nextest run`. The nextest call appears to take mostly the same arguments as cargo does, which is why I got away with placing my cargo extra arguments (`--feature <xyz>`) in `cargoNextestExtraArgs` without any issue.

If any `cargoExtraArgs` (or `llvm-cov`) is placed between `$ cargo <here> nextest run` it results in an invalid command structure. I figure both can be removed from `craneLib.cargoNextest` and the corresponding documentation without losing functionality.

I'll gladly do the work if I haven't missed something.

ipetkov:
Hi @skarmux thanks for the report!

I think I had missed that part of the documentation when writing the original integration (I suppose I mistakenly assumed just because cargo supports options like `cargo <here> build <etc>`, then `nextest` would do...), so if that's the right way to make things work I'd definitely be happy to review a PR to get it fixed!

skarmux:
Okay, this will be my first ever PR. It's assigned to this issue. I've ran `nix flake check` and everything went through. Also used the forked in my personal project as input and the barebones `craneLib.cargoNextest` check passed there as well.

PR #678, opened 2024-08-05
Cannot set cargoExtraArgs in craneLib.cargoNextest (#675)

skarmux:
## Motivation

To resolve [issue (#675)](https://github.com/ipetkov/crane/issues/675) with the `cargoExtraArgs` attribute in `craneLib.cargoNextest` being placed in between `cargo <HERE> nextest` or `cargo <HERE> llvm-cov nextest` where there should be no options/flags.

Cargo flags cannot be used in this context. Only `cargo nextest run/archive` flags, which contain all the flags from `cargo test` (at least) are possible.

Sources I have checked:
* [cargo llvm-cov nextest](https://github.com/taiki-e/cargo-llvm-cov/blob/f349cb8b99c874a152a7605489b539764624867a/docs/cargo-llvm-cov-nextest.txt)
* [cargo nextest run](https://github.com/nextest-rs/nextest/blob/e70b4e645a0df7922e396e2bfe30f3e85abe1106/site/help-text/run-help.txt)
* [cargo nextest archive](https://github.com/nextest-rs/nextest/blob/e70b4e645a0df7922e396e2bfe30f3e85abe1106/site/help-text/archive-help.txt)

## Checklist

- [x] updated `CHANGELOG.md`

### DIFF
--- a/lib/cargoNextest.nix
+++ b/lib/cargoNextest.nix
@@ -25,7 +25,7 @@ let
     "withLlvmCov"
   ];
 
-  mkUpdatedArgs = { cmd ? lib.optionalString (!withLlvmCov) "run", extraSuffix ? "", moreArgs ? "", withLlvmCov }: args // {
+  mkUpdatedArgs = { cmd ? "run", extraSuffix ? "", moreArgs ? "", withLlvmCov }: args // {
     inherit cargoArtifacts;
     pnameSuffix = "-nextest${extraSuffix}";
     doCheck = args.doCheck or true;
--- a/lib/cargoNextest.nix
+++ b/lib/cargoNextest.nix
@@ -8,7 +8,6 @@
 }:
 
 { cargoArtifacts
-, cargoExtraArgs ? ""
 , cargoLlvmCovExtraArgs ? "--lcov --output-path $out/coverage"
 , cargoNextestExtraArgs ? ""
 , doInstallCargoArtifacts ? true
@@ -19,7 +18,6 @@
 }@origArgs:
 let
   args = builtins.removeAttrs origArgs [
-    "cargoExtraArgs"
     "cargoLlvmCovExtraArgs"
     "cargoNextestExtraArgs"
     "partitions"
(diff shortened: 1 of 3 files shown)


skarmux:
I have made the decision to alter the creation of `checkPhaseCargoCommand` a little further and use if/else blocks to make it more obvious that `cargo nextest run` and `cargo llvm-cov nextest` are two distinct programs, although they share some wording and one is basically a wrapper for the other.

Since I'm still a noob with Nix, there might be benefits of using optional strings over if/else that I don't know of. Up for discussion, I guess. 😄

lib/cargoNextest.nix
```diff
@@ -44,10 +42,8 @@ let
     '';
 
     checkPhaseCargoCommand = ''
-      cargo ${cargoExtraArgs} \
-        ${lib.optionalString withLlvmCov "llvm-cov ${cargoLlvmCovExtraArgs}"} \
-        nextest ${cmd} ''${CARGO_PROFILE:+--cargo-profile $CARGO_PROFILE} \
-          ${cargoNextestExtraArgs} ${moreArgs}
+      ${if withLlvmCov then "cargo llvm-cov nextest ${cargoLlvmCovExtraArgs}" else "cargo nextest ${cmd}"} \
+      ''${CARGO_PROFILE:+--cargo-profile $CARGO_PROFILE} ${cargoNextestExtraArgs} ${moreArgs}
```
ipetkov, line 46: Is it possible/valid to bring back `${cargoExtraArgs}` but place them before `${cargoLLvmCovExtraArgs}`/`${cargoNextestExtraArgs}`?

The various subcommands all intentionally accept `cargoExtraArgs` to minimize the (re)configuration one might have to do between common and less common options. For example:

```nix
let
  commonArgs = {
    src = ...;
    cargoExtraArgs = "-p foo -p bar --features baz"; // Only build the specific crates/features for all operations
  };

  depsOnly = craneLib.buildDepsOnly commonArgs;

  tests = craneLib.cargoNextest (commonArgs // {
    cargoNextestExtraArgs = "--something --else"; // Don't need to repeat package or features here!
  });
in craneLib.buildPackage (commonArgs // {
  cargoArtifacts = tests;
});
```

ipetkov:
You can ignore the `build-std` failure, that was fixed in `d0c8f4ed8571c533b31c1d68e4ebc534b790cc5c` and should go away with a merge/rebase with `master`

lib/cargoNextest.nix
skarmux: Yes, it would work for all configurations out there. And I think you're right. It doesn't have to be removed.

I can confirm that every cargo `Build Command` according to [The Cargo Book](https://doc.rust-lang.org/cargo/commands/build-commands.html) shares all of the `cargo build` command line options.

As long as the library user is aware (I wasn't at first), only to use the common `cargo build` options in `cargoExtraArgs`, it is viable to put it back.

On another note: Since nextest allows for `cargo test` command line options, should `craneLib.cargoNextest` also consume `cargoTestExtraArgs` in addition to `cargoExtraArgs`? That way all cargo-related options can be placed in something like `commonTestArgs` if anyone ever needs that.

ipetkov: > On another note: Since nextest allows for `cargo test` command line options, should `craneLib.cargoNextest` also consume `cargoTestExtraArgs` in addition to `cargoExtraArgs`?

No I don't think we need to go that far. My (current) interpretation is that `cargoTestExtraArgs` would be for things specific to `craneLib.cargoTest`. I figure projects would want to use either `cargoTest` or `cargoNextest` but not both (though we can easily change this in the future so not too worried there)

ipetkov (approved):
Thanks again for reporting and fixing this!

PR #720, opened 2024-10-12
Add `cargoDocTest`

### LINKED ISSUE #719: `cargoDoctest`

`cargo-nextest` doesn't support doctests yet (https://github.com/nextest-rs/nextest/issues/16), so you need a separate derivation to run `cargo test --doc`.

However, `cargoTest` overwrites the `pnameSuffix` in the input, so I can't name the derivation `my-package-doctest` instead of `my-package-test`, and it's confusing to have that and `my-package-nextest` being built in tandem.

A `cargoDoctest` function as a shortcut for this behavior.

The ability to override `mkCargoDerivation` would be nice, but that's a bigger lift.

9999years:
## Motivation

Closes #719
* Add `cargoDocTest`
* Add note that `cargoNextest` doesn't run doctests
* Fix erroneous reference to `cargoTestArgs` (instead of `cargoTestExtraArgs`) in documentation for `cargoTest`

## Checklist

- [x] updated `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 
-* Added `cargoDocTest`.
+* `cargoDocTest` is now available as an alternative to `cargoTest` which runs
+  only doc tests.
 
 ### Fixed
 * Vendoring dependencies avoids creating malformed TOML configurations in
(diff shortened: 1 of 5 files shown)


ipetkov:
Hi @9999years, thanks for the contribution and the justification for why this is needed!

> `cargoTest` overwrites the `pnameSuffix` in the input, so I can't name the derivation `my-package-doctest` instead of `my-package-test`, and it's confusing to have that and `my-package-nextest` being built in tandem.

This is pretty much a fairly trivial wrapper for running doc tests. Feels a bit overkill to (set the precedent and) allow `cargoTest`'s `pname-suffix` to be overwritten so I'm happy to include this! If anything ever changes we can always express `cargoDocTest` in terms of `cargoTest` in the future so the added API surface area isn't much of a risk.

PR #731, opened 2024-11-02
Add `cargoFileset`

rijkvp:
## Motivation

This PR adds a `cargoFileset` function that includes the minimum files required to build. This allows using the new `lib.fileset` API for source filtering, which is much easier to use and automatically filters out empty directories (https://github.com/ipetkov/crane/pull/725).

## Checklist

- [ ] updated `CHANGELOG.md`

### DIFF
--- a/checks/default.nix
+++ b/checks/default.nix
@@ -346,6 +346,42 @@ in
 
   features = callPackage ./features { };
 
+  fileset = myLib.buildPackage {
+    src = lib.fileset.toSource {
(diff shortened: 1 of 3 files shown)


ipetkov (approved):
Thanks for the contribution @rijkvp! I reworked the filesets slightly just so that the individual pieces are available (in case someone wants to do their own thing but reuse existing pieces) but otherwise looks great!

PR #760, opened 2024-12-13
Replace %2F with / in url for branch

matthiasbeyer:
## Motivation

I found that the URL encoding that is put into the `Cargo.lock` file by cargo does not work in a nix build using crane, because crane does not decode the url in the generated source replacement `config.toml`.

FWIW, I know that this patch cannot be the actual solution to the problem, but it is a proof of concept that this is the issue.

We (@TheNeikos and I) used this repository:

https://github.com/cloudroots/erased-serde

with this branch: `feature/make_make_serializer_public` in the repo where we encountered the issue (unfortunately, that repo is still proprietary and closed source).

In the `branch` setting in the generated source replacement `config.toml` is `branch = "feature%2Fmake_make_serializer_public"` - which does not work with cargo and results in cargo failing during buildtime (unfortunately with `Temporary failure in name resolution for github.com`, which is not really helping here).

## Checklist

- [ ] updated `CHANGELOG.md`

### DIFF
--- a/lib/vendorGitDeps.nix
+++ b/lib/vendorGitDeps.nix
@@ -139,8 +139,12 @@ let
       let
         p = head ps;
         extractAttr = attr:
-          if p ? ${attr} then ''
-            ${attr} = "${p.${attr}}"
+          if p ? ${attr} then let
+            patchedAttr = if attr == "branch"
+            then builtins.replaceStrings ["%2F"] ["/"] "${p.${attr}}"
+            else "${p.${attr}}";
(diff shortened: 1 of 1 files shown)


ipetkov:
Hi @matthiasbeyer thanks for the PR! I've landed an alternative (and slightly more general) implementation of this in https://github.com/ipetkov/crane/pull/764. Please try that out and feel free to open a follow up issue if that does not work!

In the mean time will close this out

Issue #767, opened 2024-12-20
Heads-up: #[start] is being removed


RalfJung:
We are planning to remove the unstable `#[start]` attribute from the compiler, see https://github.com/rust-lang/rust/pull/134299 for rationale and further details. This repo showed up in a search for users of the attribute. For all we know, there's no good reason to use this attribute and you can do the following instead:
```rust
#![no_main]

#[no_mangle]
extern "C" fn main(argc: core::ffi::c_int, argv: *const *const u8) -> core::ffi::c_int {
  // ...
}
```

If that for some reason does not work, please let us know!

ipetkov:
Thanks for the heads up! This was only used in some test code that I had copied from elsewhere and I have no problems taking it out :smile: 

Will be fixed with https://github.com/ipetkov/crane/pull/769

PR #769, opened 2024-12-21
Drop start feature/attribute as its going away

ipetkov:
## Motivation
Fixes https://github.com/ipetkov/crane/issues/767

## Checklist

- [ ] updated `CHANGELOG.md`


kadiwa4:
Are you sure that `isize` is the right type for the `argc` param and the return value? Usually it would be `c_int`.

ipetkov:
IIRC when I checked `c_int` was an alias for `isize`. Either way, this doesn't matter too much, the tests are really checking that a custom source (no matter its contents) shows up in the right place, but we aren't really checking that things compile correctly (that's ultimately up to whomever is customizing what the dummy source the want)

PR #775, opened 2025-01-04
Add craneLib as self-reference to lib

fooker:
## Motivation
This just adds `self` as `craneLib` to the crane lib attrset.

Motivation here is to allow to `callPackage` other file and access the crane lib instance in there.

## Checklist

- [x] updated `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 * `cargoNextest` now supports passing `cargoNextestPartitionsExtraArgs` to each
   `cargo nextest` partition run.
-* Add self-reference `craneLib` to crane lib.
+* Add self-reference `craneLib` to crane lib instance.
 
 ## [0.20.0] - 2024-12-21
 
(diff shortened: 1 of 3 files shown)


ipetkov:
Hi @fooker thanks for the PR! Could you expand on an example use case of where this might be useful?

At a first glance I don't really see why it would be useful to access `craneLib` in any of the internals using `callPackage`: they can already access any attribute on `craneLib` by defining it in their own args attr set

fooker:
> Hi @fooker thanks for the PR! Could you expand on an example use case of where this might be useful?

The thing I try to do boils down to something like this:
```nix
myCraneLib = crane.mkLib ...;
foo = myCraneLib.callPackage ./foo.nix { }; 
```

and `foo.nix` being:
```nix
{ pkgs, lib, craneLib, ... }:

craneLib.buildPackage {
  src = craneLib.cleanCargoSource (craneLib.path ./.);
  ...
}
```

> At a first glance I don't really see why it would be useful to access `craneLib` in any of the internals using `callPackage`: they can already access any attribute on `craneLib` by defining it in their own args attr set

This is true, but would alter the example above to have a quite long argument list. And it would thing like `path` and `callPackage`, which I'd consider hard to reason where they are exactly coming from. Thus, I'd like to keep the prefix and make it more explicit.

Another argument is, that this does not hurt and follow the pattern that nixpkgs lib and pkgs are using.

ipetkov:
I see, thanks for clarifying, I think this is fine to include!

Issue #773, opened 2025-01-03
Toolchain included in runtime closure if it includes rust-src


adamdayan:
**Describe the bug**
If you override the toolchain with Rust version 1.83.0 or newer and you include the extension `rust-src` then panic messages from the stdlib include the full nix store path of the stdlib src package. These paths are not scrubbed by removeReferencesToVendoredSources because the rust-src isn't part of the cargoVendorDir and so the whole rust toolchain is included in the runtime closure, massively bloating the size of the package. This happens since 1.83.0 because rustc made a change to include the full stdlib path - see the section referencing [RFC 3127](https://rust-lang.github.io/rfcs/3127-trim-paths.html#changing-handling-of-sysroot-path-in-rustc) in the [changelog](https://releases.rs/docs/1.83.0/). 

**Reproduction**
This [flake.nix](https://github.com/adamdayan/crane-rust-src-repro/blob/develop/flake.nix#L33) is a minimal repro of the issue which modifies the quickstart example `flake.nix` with an overriden toolchain including rust-src. If you run `nix build .#packages.aarch64-linux.default --print-out-paths` and then `nix-store --query --reference <path to package>` you can see the package has a runtime dependency on the whole rust toolchain. If you run nix why-depends --precise <output package> <toolchain package> you can see it's due to a stdlib panic message.

adamdayan:
My best guess on how to fix this is to add a 3rd argument, `rustSrcDir`,  to `removeReferencesToVendoredSources` which points to the nix store path of the rust src if it's present in the toolchain and uses sed to scrub the hash. I'm very happy to give this a go and try to make an MR but I'd appreciate a pointer as to the right place to set `rustSrcDir`from and some confirmation that my idea isn't totally wrong. Thank you!

ipetkov:
Hi @adamdayan thanks for the report, yes if we can implement this in a way that can robustly detect the toolchain dir without any intervention by the package maintainer I'd be happy to have it included!

Couple of thoughts around this:
1. We should support opting out of removing references to the toolchain _just in case_ someone doesn't want this behavior (e.g. if `$doNotRemoveReferencesToRustToolchain` is set, we leave it as is)
2. We should try to automatically detect the toolchain. I think something like `rustc --print sysroot` ought to be enough, assuming the `rust-src` comes from the same derivation (anything more complex than that should be left to the caller to sort out)
[...]

PR #779, opened 2025-01-06
Scrub references to Rust toolchain

adamdayan:
## Motivation
This MR tries to address the [issue](https://github.com/ipetkov/crane/issues/773) of references to the Rust stdlib src slipping into packages when built using toolchains including rust-src. It adds a `removeReferencesToRustToolchain` `setupHook` which uses `rustc --print sysroot` to find the toolchain path and then scrubs the hash in the same manner as `removeReferencesToVendoredSources`. The tests for both rust-overlay and fenix pass.

Fixes https://github.com/ipetkov/crane/issues/773

## Checklist

- [x] updated `CHANGELOG.md`


dpc:
How do I know this is working? I did a whole workspace build in my project, then switched my flake to us the latest commit of this PR, I've rebuild and I see 2MB size difference:

Before:

```
> dust result/target.tar.zst.prev/
63M ┌── target.tar.zst.prev

```

After:

```
> dust result/target.tar.zst.prev/
61M ┌── target.tar.zst.prev
```

Not huge, but noticeable, so I'm wondering if this is what I should expect.

adamdayan:
hi @dpc, the [issue](https://github.com/ipetkov/crane/issues/773) this MR addresses only affects builds in which the rust-src is available. So if you build your project with `rust-overlay` or `fenix`'s `rust-src` extension as [here](https://github.com/ipetkov/crane/pull/779/files#diff-db1161c7f74cea34847fa6a9a7f6d15f25c65106dd41219260a671e0a37ea7fbR657) you should see the package size blow up massively; if you then use this MR the package size should go back down to the original size. By scrubbing the package of references to the `rust-src` nix store path, we avoid the whole toolchain being included in the runtime closure.

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 * `cargoNextest` now supports passing `cargoNextestPartitionsExtraArgs` to each
   `cargo nextest` partition run.
 * Add self-reference `craneLib` to crane lib instance.
+* Add `removeReferencesToRustToolchainHook`, which optionally scrubs the output 
+  binary of references to the Rust toolchain.
```
ipetkov, line 14:
```suggestion
* Add `removeReferencesToRustToolchainHook`, which, by default, removes references to
  the Rust toolchain from all installed binaries
  
### Changed
* `buildPackage` now includes `removeReferencesToRustToolchainHook` as a native dependency. To disable
  automatically removing references to the Rust toolchain, set `doNotRemoveReferencesToRustToolchain = true;`
```

docs/API.md
```diff
@@ -1916,6 +1916,21 @@ Any patched binaries on `aarch64-darwin` will be [signed](https://developer.appl
 `removeReferencesToVendoredSources "$out" "$cargoVendorDir"` will be run as a
 post install hook.
 
+### `craneLib.removeReferencesToRustToolchainHook`
+
+Defines `removeReferencesToRustToolchain()` which handles removing all
+references to the Rust toolchain from the installed binaries, which ensures that
+nix does not consider the binaries as having a (runtime) dependency on the toolchain.
+These references usually originate from panic messages. It takes 1 positional
+argument:
```
ipetkov, line 1925:
```suggestion
Defines `removeReferencesToRustToolchain()` which handles removing all
references to the Rust toolchain from the installed binaries, ensuring that the final
binaries do not have a (false) runtime dependency on the toolchain (usually due to
references in panic messages). The current toolchain is detected via `rustc --print sysroot`.
nix does not consider the binaries as having a (runtime) dependency on the toolchain.
This hook takes 1 positional argument:
```

docs/API.md
```diff
@@ -1916,6 +1916,21 @@ Any patched binaries on `aarch64-darwin` will be [signed](https://developer.appl
...
+1. the installation directory for the output.
+  * If not specified, the value  of `$out` will be used.
+  * If `out` is not specified, an error will be raised.
+
+**Automatic behavior:** if `doNotRemoveReferencesToVendorDir` is not set, then
+`removeReferencesToRustToolchain() "$out"` will be run as a
```
ipetkov, line 1931:
```suggestion
`removeReferencesToRustToolchain "$out"` will be run as a
```

flake.nix
```diff
@@ -136,14 +136,23 @@
                 (import (inputFromLock "rust-overlay"))
               ];
             };
-          in
+            fenix = import (inputFromLock "fenix") { inherit system; };
+            toolchain = (fenix.latest.withComponents [
+              "cargo"
+              "rust-src"
+              "rustc"
+              "rust-analyzer"
```
ipetkov, line 144: Is it possible to omit rust-analyzer here?

lib/buildPackage.nix
```diff
@@ -88,6 +89,7 @@ mkCargoDerivation (cleanedArgs // memoizedArgs // {
     # NB: avoid adding any non-hook packages here. Doing so will end up
     # changing PKG_CONFIG_PATH and cause rebuilds of `*-sys` crates.
     installFromCargoBuildLogHook
+    removeReferencesToRustToolchainHook
```
ipetkov, line 92: We should update the `buildPackage` API docs to mention this is included

lib/buildTrunkPackage.nix
```diff
@@ -103,6 +104,7 @@ mkCargoDerivation (args // {
     trunk
     wasm-bindgen-cli
     # Store references are certainly false positives
+    removeReferencesToRustToolchainHook
```
ipetkov, line 107: Same comment re API docs

lib/default.nix
```diff
@@ -130,6 +130,7 @@ let
       registryFromGitIndex = callPackage ./registryFromGitIndex.nix { };
       registryFromSparse = callPackage ./registryFromSparse.nix { };
       removeReferencesToVendoredSourcesHook = callPackage ./setupHooks/removeReferencesToVendoredSources.nix { };
+      removeReferencesToRustToolchainHook = callPackage ./setupHooks/removeReferencesToRustToolchain.nix { };
```
ipetkov, line 133:
```suggestion
      removeReferencesToRustToolchainHook = callPackage ./setupHooks/removeReferencesToRustToolchain.nix { };
      removeReferencesToVendoredSourcesHook = callPackage ./setupHooks/removeReferencesToVendoredSources.nix { };
```

lib/setupHooks/removeReferencesToRustToolchain.nix
```diff
@@ -0,0 +1,12 @@
...
+  lib,
+  pkgs,
+  makeSetupHook,
+}:
+makeSetupHook { 
+  name = "removeReferencesToRustToolchain"; 
+  substitutions = {
+    storeDir = builtins.storeDir;
+  };
+  propagatedBuildInputs = [ pkgs.ripgrep ];
```
ipetkov, line 11: Personally, I love ripgrep, but it would be nice if we can avoid the extra dependency here (most people will utilize the upstream cache, but in case they aren't or they are using a custom version of nixpkgs, it's one extra dependency to compile before their project can be run).

In this case it looks like ripgrep is being used to pull out the hash of the Rust toolchain, which we can also achieve via `grep` (which is already in the standard build environment)

lib/setupHooks/removeReferencesToRustToolchain.sh
```diff
@@ -0,0 +1,24 @@
+# Strip any references to the rust toolchain (which can slip in through stdlib panic info) to avoid 
+# including the toolchain in the runtime closure
+
+removeReferencesToRustToolchain() {
+    local installLocation="${1:-${out:?not defined}}"
+    # NOTE(adam.dayan): would it be better to explicitly point to $(rustc --print sysroot)/lib/rustlib/src so we only scrub references to
+    # the src? are there any reasons one might want a reference to the rest of the toolchain?
```
ipetkov, line 7: I'd say we don't need to worry about this hook being perfectly sophisticated for every possible use case. We should focus on removing references to the Rust toolchain's store path, and if that's not quite right for a particular project, they can disable the hook and tailor their own reference handling!

```suggestion

```

lib/setupHooks/removeReferencesToRustToolchain.sh
```diff
@@ -0,0 +1,24 @@
...
+    echo "stripping references to Rust toolchain"
+    local rustToolchainLocation rustToolchainStoreHash
+    rustToolchainLocation=$(rustc --print sysroot)
+    echo "Rust toolchain at: $rustToolchainLocation"
+    rustToolchainStoreHash=$(echo "$rustToolchainLocation" | rg "^@storeDir@/([a-z0-9]{32})" -or '$1')
+
+    find "${installLocation}" -type f -exec sed -i -E "s|@storeDir@/${rustToolchainStoreHash}([^/]*)|@storeDir@/eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee\1|g" {} +;
+    echo "stripping references done"
```
ipetkov, line 15:
```suggestion
    echo "stripping Rust toolchain references done"
```

lib/setupHooks/removeReferencesToRustToolchain.sh
```diff
@@ -0,0 +1,24 @@
...
+
+removeReferencesToRustToolchain() {
+    local installLocation="${1:-${out:?not defined}}"
+    # NOTE(adam.dayan): would it be better to explicitly point to $(rustc --print sysroot)/lib/rustlib/src so we only scrub references to
+    # the src? are there any reasons one might want a reference to the rest of the toolchain?
+    echo "stripping references to Rust toolchain"
+    local rustToolchainLocation rustToolchainStoreHash
+    rustToolchainLocation=$(rustc --print sysroot)
+    echo "Rust toolchain at: $rustToolchainLocation"
+    rustToolchainStoreHash=$(echo "$rustToolchainLocation" | rg "^@storeDir@/([a-z0-9]{32})" -or '$1')
```
ipetkov, line 12:
```suggestion
    rustToolchainStoreHash=$(echo "$rustToolchainLocation" | grep --only-matching '@storeDir@/[a-z0-9]\{32\}')
```

lib/setupHooks/removeReferencesToRustToolchain.sh
```diff
@@ -0,0 +1,24 @@
...
+
+    find "${installLocation}" -type f -exec sed -i -E "s|@storeDir@/${rustToolchainStoreHash}([^/]*)|@storeDir@/eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee\1|g" {} +;
```
ipetkov, line 14: Could probably make this simpler by replacing the whole hash and not worrying about capture groups (the odds of a hash collision across packages is pretty astronomical anyway)

Could also amortize the `sed` startup and compilation time by piping through xargs

```suggestion
    find "${installLocation}" -type f | xargs --no-run-if-empty sed -i'' "s|${rustToolchainStoreHash}|@storeDir@/eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee|g"
```

lib/setupHooks/removeReferencesToRustToolchain.sh
```diff
@@ -0,0 +1,24 @@
...
+    echo "stripping references done"
+}
+
+if [ -n "${doNotRemoveReferencesToRustToolchain-}" ]; then
+    echo "removeReferencesToRustToolchain disabled"
+else
+    postInstallHooks+=(removeReferencesToRustToolchain)
+fi
+
+
```
ipetkov, line 24:
```suggestion
```

ipetkov (commented):
Thanks for the PR @adamdayan, the overall direction looks great! Really like that you went with a separate hook, much easier to document and explain that way

Mostly have minor clean up comments, happy to touch them up myself if you don't get to them before I do!

lib/setupHooks/removeReferencesToRustToolchain.sh
adamdayan: much simpler, I've moved back to `grep`

adamdayan:
thanks for the v helpful comments @ipetkov , I think I've fixed all the outstanding issues

adamdayan:
@ipetkov looks like the MR is [failing the alt store test](https://github.com/ipetkov/crane/actions/runs/12689064145/job/35383355169?pr=779#step:6:196) with a permissions error when it tries to create the setup hook. Any idea how to fix this?

adamdayan:
ah I understand, the setup hook is still being written to "/nix/store/" which is not the store path in the alt-store test. Not sure why this is happening though

adamdayan:
actually, I think my above comment is incorrect - the alt-store `--store` flag doesn't replace `/nix/store` with the alt store path but instead puts `/nix/store` at an alternative root that is then chroot'ed into. I was able to get a vaguely similar error to CI when running `extra-tests/alt-store/test.sh` on a local linux machine - ` > cp: cannot create regular file '/nix/store/npyl4zs6nynsn94plygb6kkngp98l5nk-replaceCargoLockHook/nix-support/setup-hook': Read-only file system` but it's not the same error as in CI (which I don't think I have permissions to kick off). In my local case, it seems like the alt nix store was created with the wrong permissions and no owner. In CI, however, it's only my new removeReferencesToRustToolchain tests that fail not any of the other setup-hooks. So I'm still quite confused about what's going wrong in CI

ipetkov:
Not sure why its failing in CI, seems to pass for me locally :thinking:

ipetkov:
I also can't seem to make that regression test fail (by undoing part of the diff where it was originally introduced). Not sure if fixes in Nix have obviated it but maybe the solution is to get rid of the test if its not adding much value? I'll try to clean this up when I get some free time

adamdayan:
thanks a lot for putting the final polish on this @ipetkov, really appreciate your work on crane!

bakhtiyarneyman:
Thanks for doing this! This fixes my docker images being all gigantic.

girlbossceo:
Came here to also say that this PR has fixed my enormous OCI images as well. Went from ~450MB back to our normal ~55MB sizes.

PR #792, opened 2025-02-12
crane-utils/buildRustPackage - useFetchCargoVendor

BenChand:
## Motivation
https://github.com/NixOS/nixpkgs/blob/master/doc/release-notes/rl-2505.section.md

`rustPlatform.fetchCargoTarball` has been deprecated - replace with `useFetchCargoVendor`

Mainly tested this on some private repos which were failing after upgrading to latest version of rust (and throwing a deprecation warning).

## Checklist

- [x] updated `CHANGELOG.md`


ipetkov:
Thanks for the PR @BenChand ! While we're here, can we switch to using `rustPlatform.importCargoLock` instead? One less hash to manually wrangle that way

Issue #610, opened 2024-05-06
Cannot build proc-macro crate when using `lib.proc_macro = true` (underscore vs dash)


dsxmachina:
**Describe the bug**

I have a workspace containing a few binary, library and one proc-macro crate. However, when I want to build any package or use `nix flake check`, I get a weird build error:

```
       > error: `proc-macro` crate types currently cannot export any items other than functions tagged with `#[proc_macro]`, `#[proc_macro_derive]`, or `#[proc_macro_attribute]`
       >   --> core/macros/src/lib.rs:14:1
       >    |
       > 14 | pub fn main() {}
       >    | ^^^^^^^^^^^^^
       >
       > error: could not compile `macros` (lib) due to 1 previous error
       > warning: build failed, waiting for other jobs to finish...
       For full logs, run 'nix-store -l /nix/store/ymbh5xqg7z4cbiprgvs2i93fw93k27r7-cargo-package-deps-0.0.1.drv'.
```

Which is bogus, because the library does not include a `pub fn main() {}` at line 14.

**Reproduction**
I used the example init flake from your documentation (https://crane.dev/examples/quick-start-workspace.html) and replaced `my-cli` and `my-server` with my own binary crates from the workspace. Everything else is as it is in the example.

I traced down the error and found what was causing it. By looking at the derivation, I searched for the code inside the nix-store, that is used for the compilation:
```
nix show-derivation /nix/store/ymbh5xqg7z4cbiprgvs2i93fw93k27r7-cargo-package-deps-0.0.1.drv
```
This gave me quite a lot of output. Under the key "env/src" I found the link to the src repository:
```
"env": 
     ...
     "src": "/nix/store/knwb69fg1gqwhlrz4zbn23pv187za27m-source",
```

And inside `/nix/store/knwb69fg1gqwhlrz4zbn23pv187za27m-source` I found the basic directory structure of my workspace, but the Code has been replaced with a dummy implementation (which seems to be something that crane does, although I don't know why or how I could influence this behaviour).
Here is the code of the proc-macro library that is causing the compilation error:
```
#![allow(clippy::all)]
#![allow(dead_code)]
#![cfg_attr(any(target_os = "none", target_os = "uefi"), no_std)]
#![cfg_attr(any(target_os = "none", target_os = "uefi"), no_main)]

#[allow(unused_extern_crates)]
extern crate core;

#[cfg_attr(any(target_os = "none", target_os = "uefi"), panic_handler)]
[...]

PR #801, opened 2025-02-27
Update Internal `proc-macro` Handling to Support Cross-Compilation

### LINKED ISSUE #800: Proc-macro handling doesn't allow for nostd projects with custom macros
Hello! I'm working on a project that cross-compiles for an embedded chip, and have a custom `derive` macro that is a dependency of the embedded project. Unfortunately, the way that Crane handles proc-macro crates causes an issue where the derivation attempts to compile `proc-macro2` for my embedded target because `mkDummySrc` strips the `proc-macro=true` tag from the `Cargo.toml`.

I think what we should do instead is leave the `proc-macro` attributes in the configuration and condition the `pub fn main() {}` line on whether the project is a proc-macro crate or not. This should allow for intermediate compilation without attempting to compile all `proc-macro2` dependencies for an embedded target.

I'd be happy to submit a PR, but just wanted to get your thoughts on the matter first. Thanks for putting this great tool together!

gmarino2048:
## Motivation

Hello, this PR should fix #800. It updates the way we handle proc-macros by omitting the `main()` function from the dummy implementation of proc-macro crates. It should be backwards-compatible to work with #610, and should still allow the user to pass the `dummyrs` in through the derivation.

Fixes #800 

## Checklist

- [x] added tests to verify new behavior **I'd appreciate some help here, I'm not sure if this qualifies as a change in behavior**
- [ ] added an example template or updated an existing one **No template changes**
- [ ] updated `docs/API.md` (or general documentation) with changes **No API Changes**
- [x] updated `CHANGELOG.md`


lib/cleanCargoToml.nix
```diff
@@ -58,6 +56,8 @@ let
     # "name"              # let cargo manage targets/collisions/etc.
     # "crate-type"        # some tools may try to inspect crate types (e.g. wasm-pack), retain the
     #                     # definition to honor the project structure
+    # "proc-macro"        # If we have a proc-macro dependency in the workspace, rustc may try to
+    #                     # compile `proc-macro2` for the target system
```
ipetkov, line 60: We should probably add a `mkCleanSrcTests` case that the `proc-macro` attribute isn't touched

ipetkov (commented):
Hi @gmarino2048 thanks for the PR!

Are there any additional changes we can make to the `checks/proc-macro` crate as a regression test for this? Even if we can't easily assert what platform the dependencies were built for it would be nice to make the smoke test slightly more realistic if that helps!

lib/cleanCargoToml.nix
gmarino2048: I've fixed the tests and added a new `proc-macro` test to `cleanCargoToml` which covers all the possible `proc-macro` cases. I've also added a test which checks a build when setting `crate-type = ["proc-macro"]` and a case where we cross-compile a `proc-macro` for WASM, which should catch the instance I was seeing in #800.

ipetkov (approved):
Looks great, really nice work!

PR #812, opened 2025-03-16
Fix eval with allowAliases = false

K900:
## Checklist

- [ ] updated `CHANGELOG.md`


ipetkov:
Thank you for the fix!

PR #824, opened 2025-04-05
fix: `buildInputs` camel case

siph:
## Motivation

The devshell is broken for `end-to-end-testing`. This changes `buildInputs` from pascal to camel case.

## Checklist

- [ ] updated `CHANGELOG.md`


ipetkov (approved):
Good catch, thanks!

PR #836, opened 2025-04-28
doc: Add fileset filtering in flake checks section to FAQ/troubleshooting

### LINKED ISSUE #835: Consider adding a section to source filtering docs describing how sources are used in checks
Bumped into a small issue that I honestly had just been avoiding for the last few months. I recently added a markdown file as part of my docs using the include macro, and I also added some test frameworks that use toml and json files. This is really my own fault for not thinking the problem through thoroughly, but I thought maybe it would've been nice to have a section in the source filtering docs that describes how the derivations like `nextest` and `clippy` use the sources. After some debugging I realized that in a workspace, it didn't matter if I updated the the fileset for individual builds, the `nextest` check would still fail to find the files. Then it hit me that maybe it wasn't actually using the fileset from the build, and sure enough it was only inheriting the common args and artifacts. Adding a fileset `src = lib.fileset.toSources { ... };` to it fixed the issue. Would it be reasonable to ask to include this for posterity?

eureka-cpu:
## Motivation

Closes #835 

Adds a section to `source-filtering.md` which explains how to handle extra file sources which may be needed by flake checks.

## Checklist

- [ ] updated `CHANGELOG.md`

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -47,6 +47,7 @@
   * [Found invalid metadata files for crate error](./faq/invalid-metadata-files-for-crate.md)
   * [A git dependency fails to find a file by a relative path](./faq/git-dep-cannot-find-relative-path.md)
   * [Controlling whether or not hooks run during `buildDepsOnly`](./faq/control-when-hooks-run.md)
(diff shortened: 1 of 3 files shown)


ipetkov (approved):
Thanks for the docs improvement!

PR #856, opened 2025-06-13
Handling [patch] in Cargo workspaces

P-E-Meunier:
That directive is used in Cargo to modify a dependency tree. It is especially useful for large projects where fixing the underlying crates is impractical or impossible. Unfortunately, Crane needs special handling of this case since the Cargo.lock resulting from such workspaces has no registry source nor hash for the patched dependencies.

Note that this departs from what `cargo vendor` does, since `cargo vendor` vendors the original dependency instead of the patched one. The reason we need to do this here is that while `cargo vendor` (as used e.g. in `buildRustPackage`) is used to build entire workspaces, including the vendored deps, but also the [patch] directive when compiling, Crane builds vendored dependencies together without first knowing about the patches.

## Checklist

- [ ] updated `CHANGELOG.md`

### DIFF
--- a/lib/vendorCargoDeps.nix
+++ b/lib/vendorCargoDeps.nix
@@ -44,10 +44,14 @@ let
     );
 
   lock = args.cargoLockParsed or (builtins.fromTOML cargoLockContents);
+
+  toml = builtins.fromTOML (readFile (src + "/Cargo.toml"));
+
+  patch = lib.attrsets.foldlAttrs (acc: registry: value: acc // value) {} (toml.patch or {});
 in
 vendorMultipleCargoDeps (
   {
-    inherit cargoConfigs;
+    inherit cargoConfigs patch src;
     cargoLockParsedList = [ lock ];
     outputHashes = args.outputHashes or { };
     overrideVendorCargoPackage = args.overrideVendorCargoPackage or (_: drv: drv);
--- a/lib/vendorMultipleCargoDeps.nix
+++ b/lib/vendorMultipleCargoDeps.nix
@@ -42,10 +42,23 @@ in
   overrideVendorCargoPackage ? _: drv: drv,
   overrideVendorGitCheckout ? _: drv: drv,
   registries ? null,
+  patch ? { },
+  src,
 }:
 let
   cargoLocksParsed =
     (map fromTOML ((map readFile cargoLockList) ++ cargoLockContentsList)) ++ cargoLockParsedList;
+
+  linkedPatches =
+    lib.attrsets.foldlAttrs (acc: name: value: if value ? path then acc + ''
+      ln -s ${src}/${value.path} $out/${name}
+    '' else acc) "" patch;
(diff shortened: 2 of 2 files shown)


P-E-Meunier:
Absolutely, the Cargo.lock will be updated, with a path dependency, i.e. no hash, no URL. This means that if that crate is needed by others in the vendored deps (which is the point of `[patch]`: patching *transitive* dependencies, changing direct dependencies can be done directly), that crate will be filtered out by the following line:

https://github.com/ipetkov/crane/blob/2b2b564e8988e2701f8ba5a0863102c22b476bf3/lib/vendorMultipleCargoDeps.nix#L48

And hence will not be included in the vendored deps, resulting in a compilation error.

P-E-Meunier:
Absolutely! Your understanding of `[patch]` is correct. However, the whole point of `[patch]` is to replace transitive dependencies, since direct ones can be overridden directly.

When you use `[patch]`, at least one other crate in the vendored deps needs the patched version in order to compile. However, the Cargo.lock will indeed contain an entry for the local source, but that entry won't have any registry nor hash, and hence won't be included by crane, since it will be filtered by the following line:

https://github.com/ipetkov/crane/blob/2b2b564e8988e2701f8ba5a0863102c22b476bf3/lib/vendorCargoRegistries.nix#L57

This results in missing crates when pre-compiling vendored dependencies.

The tools I wrote to compile Rust in Nix, called buildRustCrate and Carnix (rewritten to Crate2nix), fixed this problem by using `cargo metadata` instead of parsing the Cargo.lock.

ipetkov:
> When you use [patch], at least one other crate in the vendored deps needs the patched version in order to compile. However, the Cargo.lock will indeed contain an entry for the local source, but that entry won't have any registry nor hash, and hence won't be included by crane, since it will be filtered by the following lines

You are correct that crates part of the workspace will _not_ be captured by the vendor step, since the intention that these are workspace sources and they should already be present in the `src` derivation and not be captured by the vendor derivation itself.

As long as you aren't filtering out the patched sources via a source filter (hence why I wanted to take a look at a reproduction at something that causes this to eliminate any potential configuration bugs or confusion) they should be present for the full build.

The only thing that comes to mind is that a patched dependency won't play nicely with source dummification. Currently we'll dummify all crates in a workspace by default on the (now seemingly flawed) assumption that local sources always consume dependencies and aren't consumed by them.

The better solution here would be to _have source dummification skip patched dependencies_ instead of patching the vendor script to include patch sources.

> The tools I wrote to compile Rust in Nix, called buildRustCrate and Carnix (rewritten to Crate2nix), fixed this problem years ago by using cargo metadata instead of parsing the Cargo.lock.

That's a fine strategy, but with `crane` we avoid needing to perform code gen or requiring IFD. Until Nix has stable/first-class support for dynamic derivations I'm not sure that using `cargo metadata` is a viable architectural decision here for us

PR #866, opened 2025-07-13
Fix Windows cross-compilation

RossSmyth:
## Motivation

Basically all non-trivial (and trivial) crates, when cross-compiled for Windows, requires pthreads. Nixpkgs and naersk have fixes for this. Here's one for Crane.

## Checklist

- [x] updated `CHANGELOG.md`


ipetkov:
Thanks for the PR @RossSmyth! I pushed some minor fixes but otherwise LGTM

PR #872, opened 2025-07-21
Add faq page about aws-lc-sys cross compilation

### LINKED ISSUE #870: Cross compiling aws-lc-sys for windows using mingw
Hi, I've been spending a significant amount of time trying to cross-compile rustls, more specifically `aws-lc-sys` with mingw for windows.

I had a special situation, because I had `aws-lc-sys` in both `dependencies` *and* `build-dependencies`. This causes issues when cross compiling, as either the build-builder or the target-builder picks up the wrong C compiler and fails with very misleading errors. One of those mentions an old gcc bug from years ago, which has long been fixed. It happens, because it tries to compile a linux test binary using the mingw compiler, then fails to execute it.

Anyways, I found fixes for these issues. They only work with up to date versions of crane and aws-lc-sys as of time of posting!!! (For anyone stumbling upon this)

When using the windows cross compilation template from the docs,
the build failures can be fixed by putting this into the `craneLib.buildPackage` options:

```nix
let buildPlatformSuffix = pkgs.lib.strings.toLower pkgs.pkgsBuildHost.stdenv.hostPlatform.rust.cargoEnvVarTarget;
in {
    CARGO_BUILD_TARGET = "x86_64-pc-windows-gnu";

    # fixes issues with aws-lc-sys
    AWS_LC_SYS_PREBUILT_NASM = 0; # just assemble it instead of using the prebuilt objects
    CFLAGS = "-Wno-stringop-overflow -Wno-array-bounds -Wno-restrict"; # ignore some warnings that pop up when cross compiling
    "CC_${buildPlatformSuffix}" = "cc";
    "CXX_${buildPlatformSuffix}" = "cc";
};
```

I create a repo with all the files necessary for reference: https://github.com/vypxl/nix-rust-windows

## Purpose of this issue

I want to share my work on this, so that others don't need to repeat it.

Maybe this could be integrated into the docs, or even baked into the nix code to allow frictionless cross compilation, as lots of projects use aws-lc-sys.

I'd be happy to do the integration part, I'd just need some pointers.

vypxl:
## Motivation

Addition of FAQ page as discussed in #870
Resolves #870

## Checklist
- [ ] updated `CHANGELOG.md`

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -48,6 +48,7 @@
   * [A git dependency fails to find a file by a relative path](./faq/git-dep-cannot-find-relative-path.md)
   * [Controlling whether or not hooks run during `buildDepsOnly`](./faq/control-when-hooks-run.md)
   * [Missing files during `checks` when filtering with filesets](./faq/missing-files-during-checks.md)
+  * [I want to cross compile aws-lc-sys to windows using mingw](./faq/cross-compiling-aws-lc-sys.md)
 ---
 * [Advanced Techniques](./advanced/advanced.md)
   * [Overriding function behavior](./advanced/overriding-function-behavior.md)
--- a/docs/faq/cross-compiling-aws-lc-sys.md
+++ b/docs/faq/cross-compiling-aws-lc-sys.md
@@ -1,0 +1,148 @@
+## I want to cross compile aws-lc-sys to windows using mingw
+
+In its C compilation process `aws-lc-sys` tries to compile small test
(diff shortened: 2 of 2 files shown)


vypxl:
Should I also add a test for this?

ipetkov:
I think this is fine as is! Adding a test means having to keep up with implementation details of aws-lc-sys and since I neither use it myself, nor do I cross compile it for windows I'm not sure I want to maintain such a test. Though happy to review docs updates if the suggested solution ever breaks!

PR #889, opened 2025-08-31
mkDummySrc: Use no_std for amdgpu

Flakebi:
## Motivation
Similar to the uefi os, the dummy src fails to compile on amdgpu when trying to use std.

## Checklist

- [x] updated `CHANGELOG.md`


ipetkov:
Thanks for the contribution!

Issue #888, opened 2025-08-31
`buildDepsOnly` rebuilds when dependencies are unchanged


adamjhf:
When using `buildDepsOnly`, dependencies are expected to only rebuild when they change. However, since #880, dependencies are now rebuilt every time an `.rs` file in the project changes, which significantly increases overall build times.

This can be reproduced using https://github.com/ipetkov/crane/tree/master/examples/quick-start-simple:
1. `nix build` builds `quick-start-simple-deps` and `quick-start-simple`
2. Modify hello world string in `main.rs`
3. `nix build` rebuilds both `quick-start-simple-deps` and `quick-start-simple`

Reverting back to previous `crane` revision works as expected:
1. Update flake input to: `crane.url = "github:ipetkov/crane/3c40c97e1881fff381e4615e82557b333edf65c4";`
2. `nix build` builds `quick-start-simple-deps` and `quick-start-simple`
2. Modify hello world string in `main.rs`
3. `nix build` rebuilds only `quick-start-simple` as expected. Dependencies are not rebuilt

This may be caused by https://github.com/ipetkov/crane/commit/25bd41b24426c7734278c2ff02e53258851db914#diff-e242285101277493d3227e4bf27fcdcafbd847aa29eddc2f84f8de09c56f3805R58, where the derivation will now change if any of `src` changes.

ipetkov:
Hi @adamjhf thanks for the report and for finding the regression!

This landed in https://github.com/ipetkov/crane/pull/880 and I totally overlooked that `srcForRemapPathPrefix = args.src or dummySrc;` will effectively leak in the path to the original source and invalidate all the dummification work.

I'm going to revert #880 for the time being since dropping `srcForRemapPathPrefix = args.src or dummySrc;` would also cause cache invalidation for anyone trying to opt _into_ the source remapping. Happy to revisit having this functionality (e.g. in a separate PR which brings things back) though if this isn't really easily reusable perhaps it makes more sense to document the fact that this exists and leave it up to projects to manually opt into source remapping on their own

cc @Popax21

PR #890, opened 2025-09-01
Revert "Automatically configure rustc source mapping for `dontStrip` builds (#880)"

ipetkov:
## Motivation

Fixes https://github.com/ipetkov/crane/issues/888

Reverts #880 for the time being since dropping `srcForRemapPathPrefix = args.src or dummySrc;` would also cause cache invalidation for anyone trying to opt _into_ the source remapping. Happy to revisit having this functionality (e.g. in a separate PR which brings things back) though if this isn't really easily reusable perhaps it makes more sense to document the fact that this exists and leave it up to projects to manually opt into source remapping on their own

cc @Popax21 

## Checklist

- [ ] updated `CHANGELOG.md`


ipetkov:
> Hmm, that's a stupid mistake on my part - sorry for missing this.

No worries I missed it too :sweat_smile: I wish there was a good way to test for this in general and have the CI keep us honest but I haven't come up with a good to actually detect it...

> I do have some more time, and I'll send you another PR to reland this feature once I did so if that's alright with you. If I don't find anything I would probably just resend the PR with the source mapping mechanism being an explicit opt-in (instead of being the default for `dontStrip`), including explicit documentation that enabling the feature breaks dependency caching - this would suck for some use cases (e.g. using Nix builds as part of a regular dev workflow with frequent rebuilds), but would still leave some other use cases intact (for example dedicated benchmarking / profiling builds for use with tools like samply).

Yes totally fine with this approach given that someone would have to explicitly opt-into it! Having all caches always invalidated for this kinda sucks, but at least it would be up to the project maintainer to decide if that's what's more important for them :+1:

PR #896, opened 2025-09-15
Always run checkPhase for cargoTest and cargoNextest

RossSmyth:
The outputs are not useful as they are just some
tarballs, so you cannot do something like
`nix run .#checks.tests`

## Motivation

If a user uses the API `cargoTest` or `cargoNextest`, they probably want to run the tests. It is also a common pattern
in nix packages to set `doCheck = false`, then with a specific check output to override it to `doCheck = true` or something
similar.

So this leads to the patterns of:
```nix
let
commonArgs = {
  blah
  blah
  blah

  doCheck = false;
};
cargoArtifacts = craneLib.buildDepsOnly commonArgs;
in {
package = craneLib.buildPackages (commonArgs // { 
  inherit cargoArtifacts;
  blah
};

tests = craneLib.cargoTest (commonArgs 
  // { 
    inherit cargoArtifacts;
  });
}
```
Which when `tests` is run, it will silently pass. So if you are like me and just always run your tests in CI, you just see the green check and are happy. But in reality the tests are not running at all.

There is no reason not to run the tests because the output of these derivations are not useful for running the tests. The outputs
are just some tarballs. If the user explicitly uses `cargoTest` or `cargoNextest`, it should run the tests. 

## Checklist

- [x] updated `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 * **Breaking**: dropped compatibility for Nix versions below 2.28.3
 * **Breaking**: dropped compatibility for nixpkgs-24.11
+* Always run tests when `cargoTest` or `cargoNextest` is used.
 * Windows cross builds provides pthreads as it is required for most crates.
 * `registryFromSparse` now passes `fetchurlExtraArgs` to the initial query for the
   registry's `config.json`
--- a/docs/API.md
+++ b/docs/API.md
@@ -659,7 +659,7 @@ environment variables during the build, you can bring them back via
 
 `cargoNextest :: set -> drv`
 
-Create a derivation which will run a `cargo nextest` invocation in a cargo
+Create a derivation which will always run a `cargo nextest` invocation in a cargo
 workspace. Note that [`cargo nextest` doesn't run
 doctests](https://github.com/nextest-rs/nextest/issues/16), so you may also
 want to build a `cargoDocTest` derivation.
@@ -771,7 +771,7 @@ environment variables during the build, you can bring them back via
 
 `cargoTest :: set -> drv`
 
-Create a derivation which will run a `cargo test` invocation in a cargo
+Create a derivation which will always run a `cargo test` invocation in a cargo
 workspace.
 
 Except where noted below, all derivation attributes are delegated to
--- a/lib/cargoNextest.nix
+++ b/lib/cargoNextest.nix
@@ -42,7 +42,7 @@ let
     // {
       inherit cargoArtifacts;
       pnameSuffix = "-nextest${extraSuffix}";
-      doCheck = args.doCheck or true;
+      doCheck = true;
 
       buildPhaseCargoCommand =
         args.buildPhaseCargoCommand or ''
--- a/lib/cargoTest.nix
+++ b/lib/cargoTest.nix
@@ -20,7 +20,7 @@ mkCargoDerivation (
   args
   // {
     inherit cargoArtifacts;
-    doCheck = args.doCheck or true;
+    doCheck = true;
 
     pnameSuffix = "-test";
     buildPhaseCargoCommand = "";


ipetkov:
Closing this in favor of https://github.com/ipetkov/crane/pull/900

dpc:
@ipetkov 

Hopefully I'm not confused about something, but IMO `cargoNextest` and similar should run `cargo nextest run` in the `build`, not `check` phase.

Yes, `check` is often used for running tests in nixpkgs. But in crane these derivations run tests as their main and only job. Running tests is not a optional "check" for them.

If crane had derivations that run unit tests in check phase, and then build the Rust project as an output, it would make sense, and `doCheck` could be used to skip the "checking" (running tests). But if someone is pulling in crane and creating chained derivations and make a `cargoNextest`-based derivation, they expect it to run the tests.

ipetkov:
> IMO `cargoNextest` and similar should run `cargo nextest run` in the `build`, not `check` phase.

@dpc I don't agree: the `check` phase is for running tests, this is also exactly where `buildDepsOnly` runs its checks (unless checks are disabled). It's intentionally done this way so that in case you want to add your own phases or hooks you can have a consistent place to do them.

Imagine a scenario where a derivation needs to perform some configuration changes, or patch things up, or maybe even disable certain tests (e.g. they can't run in the sandbox). Being able to consistently do that in a given phase/build hook across all derivations is useful (esp if check-related changes should only happen after the build phase has finished).

Changing things like `cargoTest`/`cargoNextest` to run their tests during the `build` phase could make that more complicated. I do see the point you are making, but `cargoTest`/`cargoNextest` don't just exist to run tests in isolation, I want them to fit "coherently" in terms of how the rest of the APIs behave and minimize friction elsewhere

ipetkov:
> Hmm... Makes me wonder if there is any way to compose `cargoNextest` with some other derivation, so the result (`$out`) of `cargoNextest` would be the `binaries`/`target.tar.zstd` of its input, effectively making `something -(input of)-> cargoNextest` as an input the same as `something` itself, but after running tests.

@dpc this is exactly how the APIs are currently designed today: `cargoTest` (and friends) will install their artifacts to `$out` on a successful run so they can be chained. Not saying this is something everyone would want to do, but it's there as a tool for anyone who wants to use it.

For example:

```nix
let
  commonArgs = {
    cargoArtifacts = null;
    doCheck = ...; # more on this in a bit
    # other stuff
  };

  # First build and cache deps
  depsOnly = craneLib.buildDepsOnly commonArgs;

  # Next run clippy
  clippyLints = craneLib.cargoClippy (commonArgs // {
    cargoArtifacts = depsOnly;
  });

  # Next run the tests but only after lints pass, maybe we don't want to spend any
  # CPU cycles running tests in parallel if the lints will fail
  cargoTests = craneLib.cargoTest (commonArgs // {
    cargoArtifacts = clippyLints;
  });

  # Only after the tests have passed we will actually build the package. Maybe the binaries
  # are built with fat LTO and building takes a long time and many CPU cycles and maybe
  # it's not desirable to even start running this until all the tests have passed. Maybe the test
  # suite is flaky and needs to be retried a few times, but this way we avoid having to rebuild
  # all binaries (and rerun LTO) whenever a test flakes out since that happens in a separate derivation
  myBin = craneLib.buildPackage (commonArgs // {
    cargoArtifacts = cargoTests;
    doCheck = false; # Already tested above
  });
in
  myBin
```

Now suppose with all this chaining it is later discovered that the tests don't work on a particular platform, or for certain cross-compilation scenarios and so forth. Yes you can rewrite the Nix expressions to conditionally use `clippyLints` instead of `cargoTests` in some places, but that could get a lot more difficult and annoying rather than setting something like `doCheck = stdenv.hostPlatform.isLinux && stdenv.buildPlatform.canExecute stdenv.hostPlatform;` or whatever other project-specific logic is necessary to easily disable running tests in _some_ places but not others

ipetkov:
@dpc the docs are supposed to mention that each of them delegates to `mkCargoDerivation` (that first one seems to have a typo which I'll be fixing) to avoid having to duplicate explaining the behavior.

Though feel free to propose a docs improvement which would have helped you understand this better!

ipetkov:
I mean, that's kind of what [docs](https://github.com/ipetkov/crane/blob/c5b48a59ccd5179ea626f47b05d2828c37fb31b7/docs/introduction.md) [like](https://github.com/ipetkov/crane/blob/c5b48a59ccd5179ea626f47b05d2828c37fb31b7/docs/introduction/artifact-reuse.md) [these](https://github.com/ipetkov/crane/blob/c5b48a59ccd5179ea626f47b05d2828c37fb31b7/docs/introduction/sequential-builds.md) is supposed to highlight, I'm not expecting someone to read the API _reference_ cover to cover

Anyway as always PRs welcome!

PR #901, opened 2025-09-24
Fix Windows pthreads being added to non-Windows platforms

axelkar:
## Motivation

Closes https://github.com/ipetkov/crane/discussions/884

Passes all current tests. How do I add a new one?

## Checklist

- [x] updated `CHANGELOG.md`


ipetkov:
I pushed a change to minimize the diff a bit and reformat, but otherwise looks good to me, thanks!

PR #923, opened 2025-11-17
[reland] Configure rustc source mapping

Popax21:
## Motivation
Set up a source map using [`--remap-path-prefix`](https://doc.rust-lang.org/rustc/command-line-arguments.html#--remap-path-prefix-remap-source-names-in-output) to map the builder source path (`/build/src`) to the correct source file path within the Nix store for derivations built with `dontStrip` enabled. This ensures that debuggers / other tooling can automatically find and access the correct source files. Additionally, this behavior can also be explicitly enabled/disabled using `doRemapPathPrefix`.

This is a reland of #880 which fixes issue #888 which resulted in the original PR being reverted. It does this by configuring `--remap-path-prefix` to instead remap to a "neutered" version of the `$src` Nix store path with the hash component removed - because of the way that `dummySrc` is set up this means that both `buildDepsOnly` and `buildPackage` will have the same `CARGO_BUILD_RUSTFLAGS` value, preventing unnecessary rebuilds while also not requiring that the full `$src` be passed down to the `buildDepsOnly` derivation. An additional post-build hook is used to "unneuter" any source path references again during a later part of the build process.

## Checklist
- [x] updated `CHANGELOG.md`

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -1961,6 +1961,11 @@ have to setup source mapping yourself.
    * This removes the hash part of the store path from the build fingerprint,
      which can be helpful to bypass unnecessary rebuilds resulting from usage of
      e.g. `buildDepsOnly` / `mkDummySrc`
+   * The neutered store paths are restored back to their full "unneutered"
+     counterparts using the `fixupNeuteredRustcRemapPathPrefix` setup hook,
+     which is automatically registered as a `postBuildHook` whenever
+     `configureRustcRemapPathPrefix` decides to apply neutering to a Nix store
+     path 
 
 **Automatic behavior:** runs as a pre-build hook if either `doRemapPathPrefix`
 is set to a non-empty value, or if `doRemapPathPrefix` is *not set* and
--- a/lib/setupHooks/remapPathPrefixHook.sh
+++ b/lib/setupHooks/remapPathPrefixHook.sh
@@ -1,6 +1,7 @@
 configureRustcRemapPathPrefix() {
   local remapTo="${1:-${src:?not defined}}"
   local remapFrom="${2:-$(pwd)}"
+  local doNeuter="${3:-neuter}"
 
   echo "configuring CARGO_BUILD_RUSTFLAGS to remap paths from \"${remapFrom}\" to \"${remapTo}\""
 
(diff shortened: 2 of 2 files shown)


docs/API.md
```diff
@@ -1933,6 +1934,40 @@ a temporary location defined by `$postBuildInstallFromCargoBuildLogOut`
...
+   * Otherwise, an error will be raised
+2. a path to where the source will map *from*, i.e. the current working
+   directory
+   * If not specified, the current working directory will be used
+3. `neuter` if remappings to Nix store paths should be "neutered" as part of the
+   build process (the default behavior)
+   * This removes the hash part of the store path from the build fingerprint,
+     which can be helpful to bypass unnecessary rebuilds resulting from usage of
+     e.g. `buildDepsOnly` / `mkDummySrc`
+
```
ipetkov: The docs neglect to mention that running `configureRustcRemapPathPrefix` will schedule `fixupNeuteredRustcRemapPathPrefix` to later run and fixup the sources. I think for clarity the docs should cover `fixupNeuteredRustcRemapPathPrefix` as well so its clear what happens and when it happens

lib/setupHooks/remapPathPrefixHook.sh
```diff
@@ -0,0 +1,53 @@
...
+  echo "configuring CARGO_BUILD_RUSTFLAGS to remap paths from \"${remapFrom}\" to \"${remapTo}\""
+
+  # NB: we don't want to directly remap to $src, since this will cause excess
+  # rebuilds stemming from CARGO_BUILD_RUSTFLAGS-related build fingerprint
+  # changes, as buildDepsOnly will be invoked with $src pointing to a `dummySrc`
+  # output`. Instead, we first "neuter" the path to a common intermediate by
+  # stripping the hash part of the path, which we fix up later.
+  local remapToHash=$(echo "$remapTo" | grep --only-matching '@storeDir@/[a-z0-9]\{32\}')
+  if [ "${3:-neuter}" == "neuter" ] && [ -n "$remapToHash" ]; then
+    postBuildHooks+=("fixupNeuteredRustcRemapPathPrefix \"${remapTo}\"")
```
ipetkov: Should this instead be run as a `preInstall` hook (or rather any time after checks have run, but if checks are disabled, `preInstall` is the next default phase)?

Basically what I'm wondering is whether re-patching the artifacts could lead to cargo deciding (now or in the future) to suddenly recompile the crate for the sake of testing?

lib/setupHooks/remapPathPrefixHook.sh
```diff
@@ -0,0 +1,53 @@
...
+  # https://doc.rust-lang.org/cargo/reference/config.html#buildrustflags
+  set -x
+  CARGO_BUILD_RUSTFLAGS="${CARGO_BUILD_RUSTFLAGS:-} --remap-path-prefix=${remapFrom}=${remapTo}"
+  set +x
+  export CARGO_BUILD_RUSTFLAGS
+}
+
+fixupNeuteredRustcRemapPathPrefix() {
+  local remapTo="${1:?not defined}"
+  local buildLocation="${2:-$(pwd)}"
```
ipetkov: Should this run on all of `pwd` or do we want to constrain this to only run inside of `${CARGO_TARGET_DIR:-target}`?

We have no idea what is or isn't being built as part of the rest of the derivation, and if something else is neutering its own paths we wouldn't want to incorrectly "fix them up". As a related thought, maybe instead of using `eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee` we should allocate some other (random?) invalid hash and replace it back so we're crystal clear that we aren't clobbering anything else!

docs/API.md
ipetkov: This is great, thank you for addressing! Basically I wanted to make sure we outline what things run in what phases, but we don't necessarily have to spell out every little detail

ipetkov (approved):

Issue #288, opened 2023-04-02
Consider stripping the crate version when building dependencies


ipetkov:
Originally discussed in https://github.com/ipetkov/crane/issues/281

The crux of the issue is that we automatically inherit the crate's version attribute (from `Cargo.toml`) as the version of the `buildDepsOnly` derivation. The value is also not stripped when cleaning the `Cargo.toml` file. This unfortunately means that when a crate's version is updated (e.g. to publish a new release) it will invalidate the cache and require building all dependencies from scratch.

A more cache-friendly solution would be to replace `version` with a static placeholder value (both when cleaning the `Cargo.toml` and when setting the `buildDepsOnly` derivation's version attribute).

One possible downside of this is if a project uses a build command which somehow relies on the version of the crate (e.g. tries to parse something like `cargo tree mycrate@x.y.z`). I think there should be an escape hatch (such as `doNotStripVersion = true;` or equivalent) which would bring back the previous behavior if required

ipetkov:
Some more observations on this: changing the versions of any `Cargo.toml` `[package]` definition would also need to subsequently change the respective `Cargo.lock` definitions as well, otherwise cargo will complain about the lockfile being out of date (since we pass in `--locked`)

PR #957, opened 2026-01-09
feat: make cleanCargoToml customizable

benedikt-schaber:
## Motivation

Currently `cleanCargoToml` hardcodes which paths to remove. This PR allows users to pass a `pathsModifier` function to `cleanCargoToml` to make it customizable. Further, this PR takes it from a "what to remove" to a "what to keep" approach, which seems more scalable to me. However, this means that it is a breaking change since unknown fields are now removed by default instead of kept by default.

This was motivated by me using different cargo-tools that relied in some form or another on fields filtered by `cleanCargoToml` (e.g. `cargo-leptos` uses `workspace.metadata.leptos`), which can then be easily adjusted here instead of having to roll my own `cleanCargoToml`. But could also make it easier to implement #288, I think. I could work on that if this seems like a good direction.

There are lib functions which mostly do what my helper functions do, however, these do not handle lists, while my functions do (by iterating over the children and applying the path filtering there too).

## Checklist
- [x] updated `CHANGELOG.md`

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -136,7 +136,12 @@
           inherit system;
         };
 
-        formatter = pkgs.nixfmt-tree;
+        formatter = pkgs.nixfmt-tree.override {
+          settings.excludes = [
+            "lib/filters/cargoTomlAggressive.nix"
+            "lib/filters/cargoTomlConservative.nix"
+          ];
+        };
 
         myLib = mkLib pkgs;
       in


lib/filters/cargoTomlDiscardlist.nix
```diff
@@ -0,0 +1,62 @@
+{
+  lib,
+}:
+
+let
+  # See ./cargoTomlRetainlist.nix for the fields intentionally retained with explanations.
+  # This should be the complement of that list.
```
ipetkov: If this is a complement of the other filter, could this be expressed as calling the other filter and inverting the result?

lib/cleanCargoToml.nix
```diff
@@ -1,123 +1,35 @@
...
+      builtins.map (filterData' pred path) val
+    else
+      val;
+  filterData = pred: val: filterData' pred [ ] val;
 in
 {
   cargoToml ? throw "either cargoToml or cargoTomlContents must be specified",
   cargoTomlContents ? builtins.readFile cargoToml,
+  # ([String] -> Boolean)
+  filter ? defaultFilter,
```
ipetkov: Can we name this `cargoTomlCleanFilter` for both clarity and to avoid having to rename things as they're passed down

lib/default.nix
```diff
@@ -109,6 +111,11 @@ let
       downloadCargoPackageFromGit = callPackage ./downloadCargoPackageFromGit.nix { };
       filterCargoSources = callPackage ./filterCargoSources.nix { };
 
+      filters = {
+        cargoTomlDiscardlist = callPackage ./filters/cargoTomlDiscardlist.nix { };
+        cargoTomlRetainlist = callPackage ./filters/cargoTomlRetainlist.nix { };
```
ipetkov: Got a little tripped up on these names on my first read through (in my head I swapped what's being discarded and what's being retained).

I wonder if its worth renaming these something like `cargoTomlDiscard{Conservative,Aggressive}` to better reflect what they do?

E.g. conservative tries to be careful (e.g. keep unknown values) while aggressive will try to squeeze every bit out. That way the naming highlights the _length to which we clean_ rather than trying to represent an inclusion/exclusion list (even if they're implemented that way)

docs/manifest-filtering.md
```diff
@@ -0,0 +1,91 @@
...
+to ensure that the deps are only rebuilt when really necessary.
+
+However, the `Cargo.toml` manifests can still include a lot of fields which are
+irrelevant to dependency resolution, such as documentation links, authorship
+information, and so on. To avoid rebuilds due to changes in these fields `crane`
+further filters the manifests to only include the fields relevant to dependency
+resolution using `craneLib.cleanCargoToml`.
+
+It is possible to customize this filter using `craneLib.mkDummySrc`'s
+`cargoManifestFilter` argument. `crane` provides two filters out of the box,
```
ipetkov: Didn't see any other references to `cargoManifestFilter`, probably worth revisiting this once all the renames settle down!

lib/default.nix
benedikt-schaber: Yes, that is much better.

**Why I even did it like this initially**
I thought about giving users the option to also pass their own lists e.g. to compose the default cargoTomlRetainlist with some retainlist of their own. In that case I think my naming would have made it much clearer to the user what they can do with them. However, the logic just isn't complex enough for that to make sense and so it has become an implementation detail that we indeed should not expose in the naming.

docs/manifest-filtering.md
benedikt-schaber: Good catch, I renamed the arg (`cargoTomlFilter`) but did not update the docs here. Will do

benedikt-schaber:
@ipetkov
I think I addressed your concerns with only two concerns remaining:
- `filter` argument in `cleanCargoToml`, can rename it to `cleanCargoTomlFilter` if you want. I think that is repetitive, but my opinion really is not relevant here.
- `filters.cargoTomlConservative` and `filters.cargoTomlAggressive` don't reuse code. I looked into it, I certainly could combine them into one file, but it won't safe us from listing all fields and it doesn't make sense to reuse the logic from one in the other

I would appreciate your feedback if everything else sounds good to you and feedback on these two issues.

I will look at the code and the documentation again later, probably making some further small improvement (e.g. actually replacing references in the docs with links, linking to the new doc page from other pages where relevant and perhaps using `concatMap` to save on a couple of lines of code in the filters).

PR #960, opened 2026-01-22
fix: cargoTomlAggressive only keeps top-level keys

benedikt-schaber:
## Motivation
In #957 I introduced two filters:
```
filters.cargoTomlConservative
filters.cargoTomlAggressive
```
However, pretty much all attention went to `filters.cargoTomlConservative` since you mentioned that that should be the default. Sadly, I thus overlooked, that the implementation I provided for `filters.cargoTomlAggressive` only keeps the top-level keys and so fails with any non-trivial test case.
I fixed that and ensured that now all filters are tested with all `cleanCargoTomlTests` test cases.

## Checklist
- [ ] updated `CHANGELOG.md`


benedikt-schaber:
@ipetkov I did not update the changelog, since the filters are still unreleased

PR #964, opened 2026-01-29
mkDummySrc: preserve mode and then make writeable separately

cole-h:
`--no-preserve=mode` causes `cp` to attempt to clear out the xattrs,
which involves calling the fsetxattr syscall to set the
"system.posix_acl_access" xattr.

A previous attempt used `--no-preserve-mode` + `--preserve=xattr` to
prevent needing to change any xattrs at all, but this doesn't work when
`cp` is built without xattr support. So instead we just create the effect
of `--no-preserve=mode` without using that flag: making `$out` writeable,
recursively.

Without this, builds can fail when using macOS's Virtualization.framework
to share the store into a Linux VM, because its implementation of virtiofs
doesn't seem to support xattrs.

Specifically, this fixes building crane derivations with the Native
Linux Builder.

## Checklist

- [ ] updated `CHANGELOG.md`

### DIFF
--- a/lib/mkDummySrc.nix
+++ b/lib/mkDummySrc.nix
@@ -360,7 +360,7 @@ in
 # xattrs at all, but this doesn't work when cp is built without xattr support. So instead we just
 # create the effect of --no-preserve=mode without using that flag: making $out writeable,
 # recursively.
-# 
+#
 # Without this, builds can fail when using macOS's Virtualization.framework to share the store into
 # a Linux VM, because its implementation of virtiofs doesn't seem to support xattrs.
 runCommand sourceName { } ''


lib/mkDummySrc.nix
```diff
@@ -358,7 +358,8 @@ let
 in
 runCommand sourceName { } ''
   mkdir -p $out
-  cp --recursive --no-preserve=mode,ownership ${cleanSrc}/. -t $out
+  cp --recursive --no-preserve=ownership ${cleanSrc}/. -t $out
+  chmod +w -R $out
```
dpc: The exact effect of `+w` seems to be umask-dependent. Is that OK? Should it use more explicit way of setting the permissions?

ipetkov: Looking at the generic stdenv setup (what `runCommand` eventually delegates to) it doesn't seem like `umask` is set (to something different than its default) so I don't imagine it would affect things here?

Either way if we get this wrong builds will fail if we try to clobber the files. Once the derivation finishes the write permissions will get stripped from the store path

PR #974, opened 2026-02-09
Ability to normalize workspace member packages' versions for dependency derivation

### LINKED ISSUE #445: unnecessary rebuild of unrelated packages in monorepo on member version bump
Suppose I have `crates/foo` and `crates/bar` in a monorepo. Whenever I bump the version of `crates/foo`, the content of  `Cargo.lock` at the repo root also changes. That inevitably triggers rebuild of `crates/bar` and all of its deps, even if `crates/bar` does not depends on `crates/foo` and none of its files (except for the shared `Cargo.lock`) change.

Is there any way to avoid this?

vypxl:
## Motivation

Resolves #445 

I've come across the same problem as described in the mentioned issue, and for us, it is causing many wasted CI build minutes. That's why I set out to fix it. 

This adds a new flag 'doStripVersion', which can be used to set the versions of workspace crates to a placeholder value. Doing this in the Cargo.lock and Cargo.toml's means we can reuse the dependency cache when we bump versions of our workspace crates.

## Checklist
- [x] updated `CHANGELOG.md`

### DIFF
--- a/checks/default.nix
+++ b/checks/default.nix
@@ -979,6 +979,16 @@ onlyDrvs (
           touch $out
         '';
 
+      compilesFreshSimpleStripVersion = self.compilesFresh "simple" (myLib.cargoBuild) {
+        src = ./simple;
+        doStripVersion = true;
+      };
+
+      compilesFreshStripVersion = self.compilesFresh "strip-version-test" (myLib.cargoBuild) {
+        src = ./stripVersion/sources;
+        doStripVersion = true;
+      };
+
       windows = myLibWindows.buildPackage {
         strictDeps = true;
         pname = "windows";


vypxl:
I was able to modify our production flake to fix the weird partial caching, it was a mismatch of env variables and flags, together with missing feature unification. Now the dependencies are being reused fully, with the version stripping enabled.

lib/buildDepsOnly.nix
```diff
@@ -48,6 +50,11 @@ let
   argsMaybeDummySrcOverride = if args ? dummySrc then args // { src = args.dummySrc; } else args;
 
   crateName = crateNameFromCargoToml argsMaybeDummySrcOverride;
+
+  # Use placeholder version when stripping versions to improve cache hit rates
+  versionPlaceholder = "0.0.0";
```
ipetkov: Thoughts on making the placeholder string something explicit like "version-placeholder" or even just a blank string?

It might be confusing to folks seeing logs and derivations with version `0.0.0` instead of their actual version, but seeing something named like `*placeholder*` might be a better clue

As currently written this functionality is opt-in (so using `0.0.0` is reasonable!) but if we ever decided to make this on-by-default the placeholder may matter slightly more

lib/buildDepsOnly.nix
```diff
@@ -48,6 +50,11 @@ let
...
+  effectiveVersion =
+    if doStripVersion then versionPlaceholder else (args.version or crateName.version);
```
ipetkov: Instead of explicitly handling this here (and in all other callers which set version), thoughts about extending `crateNameFromCargoToml` to understand if `args.doStripVersion` is set and have _it_ inject the placeholder? Might simplify things so callers don't need to juggle it themselves

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 * `mkDummySrc` can now pass the `cleanCargoTomlFilter` argument to `cleanCargoToml`
 * `craneLib.filters` exposes `cargoTomlAggressive`, `cargoTomlConservative`, and
   `cargoTomlDefault` for composition of custom filters for `cleanCargoToml`
+* `doStripVersion`: a new option for `buildDepsOnly`, `buildPackage`,
+  `buildTrunkPackage`, `cleanCargoToml`, and `mkDummySrc` which strips version
+  information from workspace crates in the dummy source. This improves cache hit
+  rates when only the crate version changes but dependencies remain the same.
```
ipetkov: Just leaving a note for later, the changelog entry will need to move under the `Unreleased` (after a merge or rebase with `master`) since there was a new release that went out

checks/default.nix
ipetkov: This set of tests is now failing and may need to be updated. You can iterate on it directly via `nix build -L --keep-going ./test#checks.x86_64-linux.mkDummySrcTests`, or you can run the entire test suite via `./test.sh`

lib/cleanCargoToml.nix
```diff
@@ -2,6 +2,16 @@
...
 }:
+
+{
+  cargoToml ? throw "either cargoToml or cargoTomlContents must be specified",
+  cargoTomlContents ? builtins.readFile cargoToml,
+  # ([String] -> Boolean)
+  cleanCargoTomlFilter ? filters.cargoTomlDefault,
+  doStripVersion ? false,
+  versionPlaceholder ? "0.0.0",
+}:
```
ipetkov: Just curious, why did this get moved? It was intentionally kept lower down the file as everything up to the `filterData` declaration does not depend on the actual inputs to `cleanCargoToml` (i.e. Nix can optimize not needing to re-evaluate all that on each invocation)

if we need a new `let ... in ...` binding for `cleaned` we can easily add it in below!

lib/cleanCargoToml.nix
```diff
@@ -25,11 +35,16 @@ let
...
+
+if doStripVersion && builtins.hasAttr "package" cleaned then
+  cleaned
+  // {
+    package = cleaned.package // {
+      version = versionPlaceholder;
+    };
+  }
+else
+  cleaned
```
ipetkov: This can probably be simplified with a mix of `lib.recursiveUpdate` and `lib.optionalAttrs`. Something like:

```nix
lib.recursiveUpdate cleaned (lib.optionalAttrs (doStripVersion && cleaned ? package) {
  package.version = versionPlaceholder;
})
```

lib/mkDummySrc.nix
```diff
@@ -327,9 +326,31 @@ let
...
+        pkg: if pkg ? source then pkg else pkg // { version = versionPlaceholder; }
+      ) (cargoLockContents.package or [ ]);
+    };
+
   # Since we allow the caller to provide a path to *some* Cargo.lock file
   # we include it in our dummy build only if it was explicitly specified.
-  copyCargoLock = if cargoLock == null then "" else "cp ${cargoLock} $out/Cargo.lock";
+  copyCargoLock =
+    if cargoLock == null && !builtins.pathExists (origSrc + "/Cargo.lock") then
+      null
```
ipetkov: This is probably what's causing the evaluation failure above. If there's no cargoLock file we should be interpolating an empty string (i.e. nothing to do) rather than a null here.

In general you can run the test suite via `nix flake check ./test#` (tests are defined in a sub flake for reasons beyond this PR), or you can use `./test.sh` which will run the full, comprehensive test suite

PR #975, opened 2026-02-11
downloadCargoPackageFromGit: use `cargo package` for creating the file list

### LINKED ISSUE #962: Incorrect include/exclude behavior with git dependencies (doesn't match `cargo package`)
**Describe the bug**

Package with a git dependency that has a really complicated `include` section in Cargo.toml doesn't build, because `ripgrep` excludes files that `cargo package` would leave included. (There are two conflicting rules, which Cargo and ripgrep seem to interpret differently; cargo allows the more specific include rule to take precedence over the more vague exclusion rule, while ripgrep excludes it based on the vague wildcard rule).

**Reproduction**

```console
$ nix build -L github:vikanezrimaya/crane-git-dep-include-bug
```

**Additional notes**
Adding `--debug` to the `rg` invocation in `lib/downloadCargoPackageFromGit.nix` confirmed the specific `rg` interpretation described above, while running `cargo package` in the git checkout of the aforementioned dependency and inspecting the `.crate` files confirmed that Cargo's behavior matches my expectations.

Perhaps it would be better to simply run `cargo package` in the crate checkout and then unpack the crates into `$out` instead of emulating Cargo's behavior using ripgrep?

vikanezrimaya:
## Motivation

This avoids pitfalls with `rg` misinterpreting include/exclude rules, because it uses THE SAME logic `cargo` uses for .crate files.

Because it uses Cargo itself.

Issues with lock file generation are side-stepped by using `--exclude-lockfile`, which I verified completely disables the check that the lockfile is up-to-date.

Fixes #962

## Checklist

- [x] updated `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 * `mkDummySrc` can now pass the `cleanCargoTomlFilter` argument to `cleanCargoToml`
 * `craneLib.filters` exposes `cargoTomlAggressive`, `cargoTomlConservative`, and
   `cargoTomlDefault` for composition of custom filters for `cleanCargoToml`
+* `downloadCargoPackageFromGit` now uses `cargo package -l` for generating the file
+  list, resulting in more accurate include/exclude rule interpretation. This fixes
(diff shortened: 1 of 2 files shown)


ipetkov:
Merged in https://github.com/ipetkov/crane/pull/976, thanks again!

PR #1009, opened 2026-04-04
examples: add Windows example with SQLite and Windows example without rust-overlay

axelkar:
Also added `sqlx::migrate!()` invocations.

Note that cross-compiling SQLx with macros AND native-tls doesn't currently work with Nixpkgs. The reason is that `sqlx-macros` is built for the host platform (Linux on NixOS), and depends on OpenSSL on Linux, and ~Nixpkgs doesn't seem to support having libraries in `pkgsBuildBuild`~ (Edit: seems to work with Meson? maybe it's because rust-openssl isn't using pkg-config for linking correctly??) (Edit 2: Seems that we are running into https://github.com/rust-lang/pkg-config-rs/issues/139).

## Checklist

- [ ] updated `CHANGELOG.md`


ipetkov:
Hi @axelkar thanks for the PR!

The intention behind the current set of examples is to show off the _unique_ portions for getting things up and running. Both to avoid a combinatorial explosion of recombining multiple examples together, but also to make it obvious which configuration setup is relevant to what things.

If you'd like to contribute an improvement to the examples I'd prefer that we split up the changes (e.g. the sqlx changes going to the sqlx example, and the windows cross-compilation changes going to the windows-cross compilation example)

axelkar:
Maybe a `windows-no-rust-overlay` example and a warning about macros+native-tls in the SQLx example could be added? I figure other people might be debugging the same SQLx cross-compilation issue so I'd save them the effort.

ipetkov:
FWIW I dropped `oxalica/rust-overlay` from the `cross-windows` example in https://github.com/ipetkov/crane/pull/1028 (that example was already configured to use the nixpkgs cross system so the overlay wasn't actually necessary)

I've also dropped the native-tls (openssl) option from the `sqlx` example so maybe that will help too! Though if you have any suggestions to improve that example we can iterate on it further!

axelkar:
That's probably enough. I'll close this PR.

PR #1010, opened 2026-04-06
{buildPackage,buildDepsOnly}: add `cargoBuildExtraArgs` option

musjj:
## Motivation

I wanted to pass a certain argument only to `cargo build` (and not `cargo check`, `cargo test`, etc.), but there's no option for it. There's already a `cargoCheckExtraArgs` and `cargoTestExtraArgs` option, so I've added an equivalent for `cargoBuildCommand`.

## Checklist

- [x] updated `CHANGELOG.md`


ipetkov:
Looks good thanks for the change!

PR #1019, opened 2026-04-20
mkDummySrc: Rename main as efi_main for uefi

elliotberman:
## Motivation
The default entry point for UEFI applications is "efi_main" [1]. Building the dummy source with "main" instead of "efi_main" results in a linker error when building the deps:

```
rust-lld: error: <root>: undefined symbol: efi_main
```

Add a cfg_attr to export main as efi_main so the linker succeeds.

[1]: See: https://doc.rust-lang.org/rustc/platform-support/unknown-uefi.html#requirements
> ... the entry-point is set to `efi_main` ...

## Checklist
- [x] updated `CHANGELOG.md`


ipetkov:
Thank you for the fix and the test case! Pushed up a small tweak but otherwise looks great!

PR #1030, opened 2026-04-27
Fix: Do not add a newline after the cargo-nextest args

matthiasbeyer:
When overriding the checkPhase and re-using the checkPhaseCargoCommand from the old mkCargoDerivation definition here, with all of `cargoExtraArgs`, `cargoNextestExtraArgs` and `moreArgs` empty, the trailing `\` adds a newline to the cargo-nextest call, which causes scripts to misbehave in situations like:

```nix
RUST_LOG=off ${old.checkPhaseCargoCommand} --custom-arg || somethingElse
```

With this change, this trailing newline is removed, which makes the scenario described above work.

## Checklist

- [ ] updated `CHANGELOG.md`

### DIFF
--- a/lib/cargoNextest.nix
+++ b/lib/cargoNextest.nix
@@ -66,8 +66,7 @@ let
         ${
           if withLlvmCov then "cargo llvm-cov nextest ${cargoLlvmCovExtraArgs}" else "cargo nextest ${cmd}"
         } \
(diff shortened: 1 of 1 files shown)


ipetkov:
The newline should have been escaped but maybe something gets confused with the nesting in the `''` string but thanks for the change regardless!

PR #1045, opened 2026-06-25
vendorCargoRegistries: avoid escaping shell args

llakala:
## Motivation
Using the [evaluation profiler](https://nix.dev/manual/nix/latest/advanced-topics/eval-profiler) showed me that this was a surprisingly hot call. I believe all of these uses are unnecessary. 

The result of `vendorCrate p` should just be a store path, and the result of `escapeShellArg name` is known info, since there's only one name (crates.io). `escapeShellArg "${p.name}-${p.version}"` could _technically_ be an issue, so if it's preferred, I can drop this one - but it would require a horrible name and version to trigger.

## Checklist

- [ ] updated `CHANGELOG.md`

### DIFF
--- a/lib/vendorCargoRegistries.nix
+++ b/lib/vendorCargoRegistries.nix
@@ -25,7 +25,6 @@ let
   inherit (lib)
     concatMapStrings
     concatStrings
-    escapeShellArg
     flatten
     foldl
     groupBy
@@ -65,7 +64,7 @@ let
     runCommandLocal "vendor-registry" { } ''
       mkdir -p $out
       ${concatMapStrings (p: ''
-        ln -s ${escapeShellArg (vendorCrate p)} $out/${escapeShellArg "${p.name}-${p.version}"}
+        ln -s ${vendorCrate p} $out/${p.name}-${p.version}
       '') packages}
     '';
 
@@ -158,7 +157,7 @@ let
         hashed = hash prefixedUrl;
       in
       ''
-        [source.${escapeShellArg name}]
+        [source.${name}]
         registry = "${url}"
         replace-with = "nix-sources-${hashed}"
       ''


lib/vendorCargoRegistries.nix
```diff
@@ -65,7 +64,7 @@ let
     runCommandLocal "vendor-registry" { } ''
       mkdir -p $out
       ${concatMapStrings (p: ''
-        ln -s ${escapeShellArg (vendorCrate p)} $out/${escapeShellArg "${p.name}-${p.version}"}
+        ln -s ${vendorCrate p} $out/${p.name}-${p.version}
```
ipetkov, line 67: 1. `vendorCrate p` should be a store path yes, but can store paths contain spaces? If not the `/nix/store/<hash>-<name>` path, what about subdirectories? Technically the caller can override the download derivation with whatever they want. Maybe its their own fault if they do a shell injection, but escaping here is a foolproof way of dealing with quoting
2. I'd strongly prefer we keep escaping `${p.name}-${p.version}`. There's probably a number of different ways to [set them to something potentially evil](https://crates.io/crates/cursed-trying-to-break-cargo/18446744073709551615.18446744073709551615.18446744073709551615---gREEAATT-.-V3R510N+--W0W.1.3-.nice)

lib/vendorCargoRegistries.nix
```diff
@@ -158,7 +157,7 @@ let
         hashed = hash prefixedUrl;
       in
       ''
-        [source.${escapeShellArg name}]
+        [source.${name}]
```
ipetkov, line 160: The original motivation for escaping `name` here was to avoid potential quoting issues, though its probably the wrong tool for the job anyway. I suppose these are directly controlled by the caller (via whatever registries they have configured directly in the `craneLib` instantiation) so I'm okay with relaxing this

lib/vendorCargoRegistries.nix
llakala: 1. How about we just always quote? `ln -s '${vendorCrate }'`. Depends on whether subdirectories can contain `'`.
2. I can live with that. Although it would need single quotes for escapeShellArg to do anything, which your example doesn't use.

ipetkov: If `overrideVendorCargoPackage` returns a quoted path quoting it again might end up "undoing" the quotes

PR #1057, opened 2026-07-13
fix: speed up conservative Cargo TOML filtering

mtibben:
## Motivation

`cargoTomlConservative` currently tests each visited Cargo TOML path against every removed prefix using `builtins.any` and `lib.lists.hasPrefix`. The set of removed paths is static, so large evaluations repeatedly perform the same list-prefix work.

Build a nested attribute index once and traverse it for each candidate path. Lookup cost becomes proportional to path depth rather than the number of removal prefixes, while preserving the existing filter contract for empty paths, unknown paths, exact prefixes, and descendants.

In a downstream evaluation that invokes the filter 91,720 times, this removes roughly 5.67 million prefix comparisons.

## Checklist
- [x] updated `CHANGELOG.md`


ipetkov:
Great optimization, thanks!

